Issue #17, opened 2020-07-23
Set up PR check to make sure no changes are merged which need go mod tidy


sleshchenko:
It's pretty easy to forget to do `go mod tidy` after you change go.mod.
If we set up PR check that will do `go mod tidy` and check if not changes are made in go.sum
then we make sure that go.sum has actual state after each PR.

sleshchenko:
P.S. `go fmt -x ./...` could be performed as well to make sure that code does not need formatting.

PR #20, opened 2020-07-23
Adding git actions

### LINKED ISSUE #11: Setup Travis CI to validate PRs
A minimum " go build ... " would be a good starting point

ranakan19:
added GitActions to do the following:
- get dependencies
- check build
- run tests
- check format
- check go mod status

Closes #11 and #17

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,54 @@
+name: Go
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.13
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
+            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+            dep ensure
+        fi
+
+    - name: Build main
+      run: go build main.go
+
+    - name: Run Go Tests
+      run: go test ./...
+    
+    - name: Check format
+      run: |
+        go fmt -x ./...
+        if [[ ! -z $(git status -s) ]]
+        then
+          echo "not well formatted sources are found"
+          exit 1
+        fi
+    - name: Check go mod status
+      run: |
+        go mod tidy
+        if [[ ! -z $(git status -s) ]]
+        then
+          echo "Go mod state is not clean"
+          exit 1
+        fi
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -22,13 +22,6 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
-    - name: Build main
-      run: go build main.go
-    
-    # Commenting tests untill tests are fixed, mentioned in Issue#18
-    #- name: Run Go Tests
-    #  run: go test ./...
-
     - name: Check go mod status
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
+            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+            dep ensure
```
sleshchenko, line 30: As far as I see that project uses go modules but not `go dep`... Could you elaborate on why then it's here?

.github/workflows/go.yml
```diff
@@ -0,0 +1,54 @@
...
+        fi
+
+    - name: Build main
+      run: go build main.go
+
+    - name: Run Go Tests
```
sleshchenko, line 36: Tests are currently broker and I think it's better to disable that action until they are not fixed in https://github.com/devfile/parser/issues/18
```suggestion
      # tests are disabled since the are broken
      # will be fixed in https://github.com/devfile/parser/issues/18
    - if: false
      name: Run Go Tests
```

.github/workflows/go.yml
```diff
@@ -0,0 +1,54 @@
...
+      run: go test ./...
+    
+    - name: Check format
```
sleshchenko, line 39: It's not obvious but as far as I see now - it's better to check go.sum first and them `go fmt` because `go fmt` could also add missing dependencies to go.sum. See https://github.com/devfile/parser/runs/905595388?check_suite_focus=true

wtam2018:
> parser (binary file) got committed. Perhaps, we can create GiHub Actions to check for binary files
Add "parser" to .gitingore should be good enough.

sleshchenko:
> Add "parser" to .gitingore should be good enough.

The question: how that binary was built? If it's needed to manually type a command and specify binary name - anyone are free to name binary as they wish.
If we want to name binary named like parser, or devfile-parser (may be precise a bit) then we can set up Makefile rule to compile with binary with a standardized name which is in gitignore. But it's a bit out of the scope of the current PR.

.github/workflows/go.yml
ranakan19: Agreed

.github/workflows/go.yml
wtam2018: It is probably because it is in the default go workflow action.  It shouldn't harm anything.  Though.

sleshchenko: Agree that it shouldn't harm anything because it even checks if Gopkg.toml exist which does not in our case. So, it's a bit redundant and I would vote for removing if there is no reasons to keep it.

wtam2018: I am fine with removing it.

wtam2018:
> > Add "parser" to .gitingore should be good enough.
> 
> The question: how that binary was built? If it's needed to manually type a command and specify binary name - anyone is free to name binary as they wish.
> If we want to name binary named like parser, or devfile-parser (may be precise a bit) then we can set up Makefile rule to compile with binary with a standardized name which is in gitignore. But it's a bit out of the scope of the current PR.

`go build` will do it.   It looks like it the name is picked up from the go module name.

.github/workflows/go.yml
wtam2018: +1 I ran into that as well.

.github/workflows/go.yml
ranakan19: I'm fine removing it as well, kept it as it was default suggestion for a github action for go.
Removed it in 9fb6014

.github/workflows/go.yml
ranakan19: commented test target in [9fb6014](https://github.com/devfile/parser/pull/20/commits/9fb6014609641dc7e3303be4795b04b3ece55289)

PR #22, opened 2020-07-24
Git to ignore "parser" binary file

wtam2018:
This PR add "parser" (the compiled binary) to .gitignore and remove "main" as it is the default binary built when "go build" is executed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.so
 *.dylib
 parser
+main
 
 # Test binary, built with `go test -c`
 *.test
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 *.dll
 *.so
 *.dylib
-main
+parser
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -4,7 +4,7 @@
 *.dll
 *.so
 *.dylib
-main
+parser
```
sleshchenko, line 7: The default binary name is still main, right?

If so - then please consider adding a tiny Makefile rule to compile binary with a wanted name.
If we take into account that binary could be moved outside of the project folder - then devfile-parser could be better name. WDYT?

wtam2018: If you do a `go build`, `parser` gets built.   But if you do `go buld main.go`,  `main` is produced.  I guess I should add back `main`.    I'd agree that we should create a Makefile.  In this PR, the goal is to stop binaries getting merged inadvertently.

PR #27, opened 2020-08-04
use devfile/kubernetes-api structs

ranakan19:
### What does this PR do?
- replaces in-place defined structs with the common structs defined in devfile/kubernetes-api
- removes devfile 1.0.0 support

in references to  devfile/parser#26, WIP

yes, fixed existing tests.
1. ``` make ``` - builds main.go
2. ``` make test``` to check tests
3. ``` ./main``` should be able to read from components - dockerfile, containers and commands

### Dependencies
- Dockerfile component needs to be added on kubernetes-api, and consequently removed from devfile/parser


main.go
```diff
@@ -15,7 +14,15 @@ func main() {
 		for _, component := range devfile.Data.GetAliasedComponents() {
```
adisky, line 14: `GetAliasedComponents()`, this method could be removed, It was required for devfile V1, for V2 name is a required field

pkg/devfile/parser/data/2.0.0/components.go
```diff
@@ -27,12 +28,12 @@ func (d *Devfile200) GetCommands() []common.DevfileCommand {
 }
 
 // GetParent returns the  DevfileParent object parsed from devfile
-func (d *Devfile200) GetParent() common.DevfileParent {
+func (d *Devfile200) GetParent() apiComp.Parent {
```
adisky, line 31: why apiComp?  how about `v1.Parent`?

pkg/devfile/parser/data/2.0.0/types.go
```diff
@@ -23,388 +15,17 @@ type Devfile200 struct {
 	Components []common.DevfileComponent `json:"components,omitempty"`
 
 	// Bindings of commands to events. Each command is referred-to by its name.
-	Events common.DevfileEvents `json:"events,omitempty"`
+	Events apiComp.WorkspaceEvents `json:"events,omitempty"`
 
 	// Optional metadata
 	Metadata common.DevfileMetadata `json:"metadata,omitempty"`
```
adisky, line 21: why some fields are still referring to `common`?

main.go
ranakan19: Thank you, removing the function

pkg/devfile/parser/data/2.0.0/components.go
ranakan19: yes, v1 seems better, will switch to v1.

ranakan19: resolved in 679c280

pkg/devfile/parser/data/2.0.0/types.go
ranakan19: Dockerfile component is to be part of devfile 2.1.0 schemaVersion, and since devfile/api doesn't have dockerfile component added yet, it is defined within the library. Thus the components are still wrapped as DevfileComponent defined in common to support both 2.0.0 and 2.1.0 from the library, untill dockerfile component is added on the devfile/api and supports other 2.1.0 components.

pkg/devfile/parser/data/2.0.0/components.go
```diff
@@ -3,6 +3,7 @@ package version200
 import (
 	"strings"
 
+	v1 "github.com/devfile/kubernetes-api/pkg/apis/workspaces/v1alpha1"
```
wtam2018, line 6: It is really "v1alpha1" not "v1".

pkg/devfile/parser/data/versions.go
```diff
@@ -29,7 +29,7 @@ var apiVersionToDevfileStruct map[supportedApiVersion]reflect.Type
 // Initializes a map of supported devfile api versions and devfile structs
 func init() {
 	apiVersionToDevfileStruct = make(map[supportedApiVersion]reflect.Type)
-	apiVersionToDevfileStruct[apiVersion100] = reflect.TypeOf(v100.Devfile100{})
+	// apiVersionToDevfileStruct[apiVersion100] = reflect.TypeOf(v100.Devfile100{})
```
wtam2018, line 32: Remove it instead of comment it out.

PR #34, opened 2020-10-15
Port over odo parser

yangcao77:
### What does this PR do?
This PR moves the parser pkg from Odo repo to devfile/parser, and integrates with latest devfile go structs defined in devfile/api

Fixes https://github.com/devfile/api/issues/183

Tested by running the unit tests within this PR.


main.go
```diff
@@ -41,10 +42,10 @@ func main() {
...
-		devfile, err = parser.ParseAndValidate(devfileLocation)
-		return devfile, err
-	*/
-	return
+
+	var devfile parser.DevfileObj
+	devfile, err = devfilepkg.ParseAndValidate(devfileLocation)
+	return devfile, err
+
+	// return
```
amisevsk, line 50:
```suggestion
```

main.go
```diff
@@ -41,10 +42,10 @@ func main() {
 
 //ParseDevfile to parse devfile from library
 func ParseDevfile(devfileLocation string) (devfileoj parser.DevfileObj, err error) {
```
amisevsk, line 44: Named returns aren't used here (`devfileoj`). We should either have
```go
func ParseDevfile(devfileLocation string) (devfile parser.DevfileObj, err error) {
//...
devfile, err = devfilepkg.ParseAndValidate(devfileLocation)
```
or
```go
func ParseDevfile(devfileLocation string) (parser.DevfileObj, error) {
//...
devfile, err := devfilepkg.ParseAndValidate(devfileLocation)
```

main.go
```diff
@@ -41,10 +42,10 @@ func main() {
...
-	/*
-		var devfile parser.DevfileObj
-		devfile, err = parser.ParseAndValidate(devfileLocation)
-		return devfile, err
-	*/
-	return
+
+	var devfile parser.DevfileObj
```
amisevsk, line 46: Not necessary to have this declaration on a separate line

pkg/devfile/parse.go
```diff
@@ -0,0 +1,62 @@
+package devfile
+
+import (
+	"github.com/devfile/parser/pkg/devfile/parser"
+	// "github.com/devfile/parser/pkg/devfile/validate"
```
amisevsk, line 5: Comments in import blocks break auto-formatting (e.g. by goimports)

pkg/devfile/parse.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// This is the top level parse code which has validation code specific to odo hence it cannot be kept inside the
+// devfile/parser package. That package is supposed to be independent of odo.
+
+// ParseFromURLAndValidate func parses the devfile data from the url
+// and validates the devfile integrity with the schema
+// and validates the devfile data.
+// Creates devfile context and runtime objects.
+func ParseFromURLAndValidate(url string) (d parser.DevfileObj, err error) {
```
amisevsk, line 15: The functions in this file are just wrappers on `parser.Parse[...]()` *without* validation. I'm not clear why they should exist (and why consumers wouldn't just do parsing and validation in two steps.

pkg/devfile/parser/configurables.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+func (d DevfileObj) GetMetadataName() string {
+	return d.Data.GetMetadata().Name
+}
+
+// FormatDevfileSupportedParameters outputs supported parameters and their description
+func FormatDevfileSupportedParameters() (result string) {
+	for _, v := range GetDevfileSupportedParameters() {
+		result = result + " " + v + " - " + supportedDevfileParameterDescriptions[v] + "\n"
```
amisevsk, line 189: Consider using `fmt.Sprintf` here and a `[]String`, e.g.
```go
result = []String{"Available Parameters for Devfile Components:"}
for _, v := range GetDevfileSupportedParameters() {
  result = append(result, fmt.Sprintf("%s - %s", v, supportedDevfileParameterDescriptions[v]))
}
return strings.Join(result, "\n")
```

pkg/devfile/parser/data/common/command_helper.go
```diff
@@ -0,0 +1,71 @@
...
+	return dc.Id
+}
+
+// SetIDToLower converts the command's id to lower case for more efficient processing and returns the new id
+func SetIDToLower(dc *v1.Command) string {
+	var newId string
+	newId = strings.ToLower(dc.Id)
+	dc.Id = newId
+	return newId
+}
```
amisevsk, line 20: What's the difference between this and just inlining
```go
command.Id = strings.ToLower(command.Id)
```
where we would use this function?

pkg/devfile/parser/data/common/command_helper.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"strings"
+
+	v1 "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
+)
+
+// GetID returns the ID of the command
+func GetID(dc v1.Command) string {
+	return dc.Id
```
amisevsk, line 11: Since `dc.Id` is exported, a getter is not needed (unless there's a plan to do some actual processing here in the future).

pkg/devfile/parser/data/common/component_helper.go
```diff
@@ -0,0 +1,25 @@
+package common
+
+import (
+	v1 "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
+	"k8s.io/klog"
+)
+
+// IsContainer checks if the component is a container
+func IsContainer(component v1.Component) bool {
```
amisevsk, line 9: Could these functions be replaced with one?
```go
func getComponentType(component v1.Component) v1.ComponentType
```
and then usages could be
```go
type := getComponentType(myComponent)
if type == v1.ContainerComponentType {//...}
// or
if type == v1.VolumeComponentType {//...}
```
This would also allow you to use the result in a `switch` statement.

pkg/devfile/parser/data/common/project_helper.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+
+	v1 "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
+)
+
+// GetDefaultSource get information about primary source
+// returns 3 strings: remote name, remote URL, reference(revision)
+func GetDefaultSource(ps v1.GitLikeProjectSource) (string, string, string, error) {
```
amisevsk, line 11: Named returns make sense here

pkg/devfile/parser/data/v2/header.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+//SetSchemaVersion sets devfile schema version
+func (d *DevfileV2) SetSchemaVersion(version string) {
+	d.SchemaVersion = version
+}
+
+// GetMetadata returns the DevfileMetadata Object parsed from devfile
+func (d *DevfileV2) GetMetadata() devfilepkg.DevfileMetadata {
+	return d.DevfileHeader.Metadata
```
amisevsk, line 14: While there are benefits to resolving the inlined struct in this way, it's probably not necessary -- we can just do
```
return d.Metadata
```
at which point this function seems unnecessary.

amisevsk (commented):
In general I think it's a bad idea to have no-logic getters on exported fields (e.g. `return d.Parent`) unless we can justify more complicated logic in the future.

pkg/devfile/parse.go
yangcao77: the validation will be uncommented once we have it ready in devfile/api

pkg/devfile/parser/data/common/component_helper.go
yangcao77: I have added `GetComponentType` function. 
IMO the `IsContainer` and `isVolume` or other component type can still be left as a helper function; it's makes consumers' life easier that they don't need to do extra check. 
WDYT?

pkg/devfile/parser/data/common/command_helper.go
yangcao77: I think you are right, except for they function has a return value which returning the new command.Id after setting it. 
this is how we use it in our code: https://github.com/devfile/parser/blob/3a759b5e0196f8cb5908054a83aa0537773874bf/pkg/testingutil/devfile.go#L84
But I'm fine with removing this function.

pkg/devfile/parser/configurables.go
yangcao77: Good catch! This function is for odo, and should be removed from the generic parser repo. 
`GetDevfileSupportedParameters()` is enough for the library

yangcao77:
@amisevsk they reason why we have those `Get()` functions is because information like `d.component` is not exposed in `devfileObj` which consumers will get after parsing.  They can only use interface defined in devfileData to access the struct. 
This is how the initial devfile parser was implemented, and is how odo currently using and implemented.

pkg/devfile/parser/data/common/component_helper.go
```diff
@@ -23,3 +24,26 @@ func IsVolume(component v1.Component) bool {
...
+	if component.Kubernetes != nil {
+		return v1.KubernetesComponentType, nil
+	}
+	if component.Openshift != nil {
+		return v1.OpenshiftComponentType, nil
+	}
+	if component.Custom != nil {
+		return v1.CustomComponentType, nil
+	}
+	return "", fmt.Errorf("Unknown component type")
```
maysunfaisal, line 48: err should not be capital. I think we can also opt for a switch case here instead of if blocks

pkg/devfile/parse.go
amisevsk: Makes sense, thanks.

pkg/devfile/parser/data/common/component_helper.go
amisevsk: Makes sense to me, especially for a library.

pkg/devfile/parser/context/fake.go
```diff
@@ -0,0 +1,10 @@
+package parser
```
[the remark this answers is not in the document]
yangcao77: all fields in`DevfileCtx` are defined as internal, and cannot be accessible outside of the context package

pkg/devfile/parser/data/common/component_helper.go
```diff
@@ -0,0 +1,47 @@
...
+
+// IsContainer checks if the component is a container
+func IsContainer(component v1.Component) bool {
+	if component.Container != nil {
+		klog.V(2).Infof("Found component \"%v\" with name \"%v\"\n", v1.ContainerComponentType, component.Name)
+		return true
+	}
+	return false
+}
+
```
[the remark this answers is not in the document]
yangcao77: These functions were port over from odo repo, and I'm fine with removing those log statements

pkg/devfile/parser/context/fake.go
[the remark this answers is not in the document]
wtam2018: You are right.  Let's move FakeContext() to xxx_test.go so that it is not visible outside of tests.   Also, we should consider not exporting this function.

pkg/devfile/parser/data/v2/common/component_helper.go
```diff
@@ -19,7 +17,6 @@ func IsContainer(component v1.Component) bool {
 // IsVolume checks if the component is a volume
 func IsVolume(component v1.Component) bool {
 	if component.Volume != nil {
-		klog.V(2).Infof("Found component \"%v\" with name \"%v\"\n", v1.VolumeComponentType, component.Name)
 		return true
 	}
 	return false
```
wtam2018: Simplify to:
```
return component.Volume != nil
```

pkg/devfile/parser/data/v2/common/component_helper.go
```diff
@@ -4,13 +4,11 @@ import (
...
-	"k8s.io/klog"
 )
 
 // IsContainer checks if the component is a container
 func IsContainer(component v1.Component) bool {
 	if component.Container != nil {
-		klog.V(2).Infof("Found component \"%v\" with name \"%v\"\n", v1.ContainerComponentType, component.Name)
 		return true
 	}
 	return false
```
wtam2018: Simplify to:
```
return component.Container != nil
```

kadel:
I don't thing that this is a good approach. 

I was under assumption that we have agreement on evolving and doing all necessary changes to parser in odo repo first, and then once we have independent parser package in odo we extract it from odo and put it here. 
If we start evolving parser package like this in a separate repo odo and this parser package will drift apart and we will never be able to switch to this. 

The first step was supposed to be in a the opposite direction. If there are additional things in devfile/parser that are not in odo it should be added into odo parser package.

wtam2018:
When will odo parser finish with evolving?

kadel:
> I don't thing that this is a good approach.
> 
> I was under assumption that we have agreement on evolving and doing all necessary changes to parser in odo repo first, and then once we have independent parser package in odo we extract it from odo and put it here.
> If we start evolving parser package like this in a separate repo odo and this parser package will drift apart and we will never be able to switch to this.
> 
> The first step was supposed to be in a the opposite direction. If there are additional things in devfile/parser that are not in odo it should be added into odo parser package.

I haven't seen this issue on odo side before https://github.com/openshift/odo/issues/4117#issue-722538678
It addresses all my concerns. I just wanted to make sure that we don't have the same logic in odo and in devfile/parser.
The [issue]( https://github.com/openshift/odo/issues/4117#issue-722538678) talks about using devfile/parser in odo.  We just have to make sure that it happens right after this PR is merged.

pkg/devfile/parser/context/fake.go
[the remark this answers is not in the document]
yangcao77: I agree we should rename the file to avoid confusion. 
The function should be kept as exported, because consumers(like odo) would need that function for their testing purpose.

johnmcollier:
Should we be vendoring the go module dependencies here? Might be a good idea

pkg/devfile/parse.go
```diff
@@ -0,0 +1,61 @@
+package devfile
+
+import (
+	"github.com/devfile/parser/pkg/devfile/parser"
+)
+
+// This is the top level parse code which has validation code specific to odo hence it cannot be kept inside the
```
adisky, line 7: why we need this here? I think it could be removed, later on we need to think how we want to expose library functions.

pkg/devfile/parser/data/v2/2.0.0/types.go
```diff
@@ -0,0 +1,5 @@
+package version200
+
+// Devfile200 is the struct for devfile 2.0.0 specific implementation
+type Devfile200 struct {
```
adisky, line 4: do we need this now? if not it can be removed

pkg/devfile/parser/types.go
```diff
@@ -3,15 +3,21 @@ package parser
 import (
 	"encoding/json"
 	"fmt"
-	"reflect"
 	"strings"
 
 	v1 "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
 	devfileCtx "github.com/devfile/parser/pkg/devfile/parser/context"
 	"github.com/devfile/parser/pkg/devfile/parser/data"
+	"github.com/pkg/errors"
```
adisky, line 11: shouldn't this file be renamed to something else pkg/devfile/parser/types.go? and also this methods need to be implemented by devfileObj `DevfileObj` ?

pkg/devfile/parser/data/v2/2.0.0/types.go
yangcao77: We intentionally add this here for our future devfile 2.1 support. @maysunfaisal

pkg/devfile/parser/types.go
yangcao77: I agree to rename the file. `types.go` is too confusing
This is how we implement in odo as well, I don't think we want to change the implementation.

adisky: yes for now lets rename the file.

pkg/devfile/parser/data/v2/2.0.0/types.go
adisky: why we need our own structure in library, i think it will complicate things @maysunfaisal anything you have in mind that i am not able to understand.

main.go
```diff
@@ -4,8 +4,9 @@ import (
 	"fmt"
 	"reflect"
 
+	devfilepkg "github.com/devfile/parser/pkg/devfile"
 	"github.com/devfile/parser/pkg/devfile/parser"
-	v200 "github.com/devfile/parser/pkg/devfile/parser/data/2.0.0"
+	v2 "github.com/devfile/parser/pkg/devfile/parser/data/v2"
```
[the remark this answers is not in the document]
yangcao77: I understood what do you mean. 
Actually this import alias was automatically added by vscode go plugin. I guess to make the imports neat and easy to view.

pkg/devfile/parser/data/v2/2.0.0/types.go
yangcao77: For minor version change of devfile schema version, they should be backward compatible, but might add additional fields in higher version. (in this case, let's say 2.1.0). They should share the same common methods defined in `/common` and `/v2`. but for 2.1.0, it might need specific functions, and might require different implementation on same function for 2.0 & 2.1. 
we define our own struct like that in order to support both 2.0 and 2.1. 
This is an example branch we used for testing, which 2.0 and 2.1 have different implementation on `GetCustomType()`: https://github.com/maysunfaisal/parser/commit/3fd444c23f259b55da551760efb4c5633585cda3

yangcao77: For now, the empty struct looks like meaning less; we just set it up for future 2.1 support.

maysunfaisal: @adisky Yes, the idea behind this is; there may be feature differences between 2.0.0 & 2.1.0. For example, 2.1.0 will be having build guidance which 2.0.0 is not going to support. Hence we laid out the dir 2.0.0 & 2.1.0 with their own structs and an  interface pkg/devfile/parser/data/v2/interface.go(empty for now) which will have the func definition GetBuildGuidance/AddBuildGuidance, etc where 2.0.0 and 2.1.0 will have their own implementation of the func.

This is exactly like the outermost `DevfileData` but one level nested to handle different implementation versions between 2.0.0 & 2.1.0. Check the New func in pkg/devfile/parser/data/v2/helper.go where we return either of the above structs to the new empty interface(which will eventually have funcs for new features in 2.1.0)

maysunfaisal: You may check this commit db34d9f4ecc1802426e7ae1962fbeb0f69e6f40f for a better idea of what the above paragraph just described

amisevsk:
IMO if we're going to start vendoring dependencies in this repo, it should be done in a separate PR.

yangcao77:
@amisevsk I'm fine with either way. I will modify the commit history if we want to remove the vendor folder. I don't want the git history to be too big. 
Is that the only concern from you?

@adisky And additional concerns?

I would like to get approval from you two before merging this PR. 
Let me know if you found difficulty reviewing with the PR due to the vendor folder. I will remove it if that's the case.

adisky:
>  I don't want the git history to be too big.

@yangcao77 To keep the history clean you can squash all the commits before merging

> @adisky And additional concerns?
> 
> I would like to get approval from you two before merging this PR.
> Let me know if you found difficulty reviewing the PR due to the vendor folder. I will remove it if that's the case.

For Porting purpose from odo to here, It is good to go from my side but more work needed to make it consumable as library 
Logged two issues in devfile/api
https://github.com/devfile/api/issues/194
https://github.com/devfile/api/issues/193

amisevsk:
If we're squashing this commit before merging then removing the vendoring is essential -- otherwise the commit that ends up in the history is completely useless from a history point of view. I have no problem with force-pushes to PR branches personally. 

Vendor vs not vendor is a topic worth discussing and not something we should slide into a PR about something generally separate.

yangcao77:
I do not want to squash all commit into one single commit, that would lose all commit history inside this PR. 
Instead, I have removed the commits related to vendoring from commit history of the PR. 
Let me know if you have any other concerns. @amisevsk

yangcao77:
I'm going to merge this PR if no additional concerns have been raised end of Tuesday(EDT). We need this PR to proceed our next step. Thanks.

maysunfaisal (approved):
All the review comments were addressed for the PR

/lgtm

PR #44, opened 2020-11-09
Add Devfile Kube Struct Library Functions

maysunfaisal:

### What does this PR do?
This PR copies over the refactored generator code from the Odo repo (https://github.com/openshift/odo/pull/4134)

Fixes https://github.com/devfile/api/issues/184


maysunfaisal:
@andrewballantyne this is the PR for the Library functions which should cover those 4 shortlisted ones from https://github.com/openshift/console/pull/6321/files#diff-f8f657472eec48f6e5d2eb27f345b88687919d40900828812362ad5148d29683R35-R38

pkg/devfile/generator/utils.go
```diff
@@ -0,0 +1,170 @@
...
+		corev1.EnvVar{
+			Name:  EnvProjectsSrc,
+			Value: syncFolder,
+		})
+
+	return nil
+}
+
+// ContainerParams is a struct that contains the required data to create a container object
+type ContainerParams struct {
```
mik-dass, line 113: Can we make this a non exported struct? It seems to be only used inside this package.

pkg/devfile/generator/generators_test.go
```diff
@@ -0,0 +1,570 @@
...
+
+			if buildConfig.Spec.Source.Git.Ref != tt.GitRef || buildConfig.Spec.Source.Git.URI != tt.GitURL {
+				t.Error("TestGetBuildConfig error - build config git source does not match")
+			}
+		})
+	}
+
+}
+
+func fakeResourceRequirements() *corev1.ResourceRequirements {
```
mik-dass, line 547: Maybe put this in the `testingutil` package with input based limits for future use.

maysunfaisal: What do you think about interfacing the exported generator functions like we discussed?

mik-dass: We can leave that for now as that would require a lot of unit test code changes from odo's side.

pkg/devfile/generator/generators.go
```diff
@@ -0,0 +1,327 @@
...
+			IsPrivileged: false,
+			Command:      comp.Container.Command,
+			Args:         comp.Container.Args,
+			EnvVars:      envVars,
+			ResourceReqs: resourceReqs,
+			Ports:        ports,
+		}
+		container := getContainer(containerParams)
+
+		// If `mountSources: true` was set, add an empty dir volume to the container to sync the source to
```
adisky, line 66: Are we sure that all projects needs this? emptyDir volume?

pkg/devfile/parser/data/v2/components.go
```diff
@@ -10,6 +10,52 @@ func (d *DevfileV2) GetComponents() []v1.Component {
...
+		if comp.Volume != nil {
+			components = append(components, comp)
+		}
+	}
+	return components
+}
+
+// GetPortExposure iterate through all endpoints and returns the highest exposure level of all TargetPort.
+// exposure level: public > internal > none
+func (d *DevfileV2) GetPortExposure() map[int]v1.EndpointExposure {
```
adisky, line 37: not sure about this one in this package? is it intended to be directly consumed by library users? i saw  generators code using this

adisky (commented):
@maysunfaisal Mostly looks good to me, just had two concerns.

pkg/devfile/generator/generators.go
maysunfaisal: This is just a comment, we are not doing any empty volume mounting. If mountSources is true, we just set `PROJECTS_ROOT` & `PROJECT_SOURCE` for the container.

I'll go through the comments to make sure they're updated and clean.

pkg/devfile/parser/data/v2/components.go
maysunfaisal: True, it is used by the generator code atm. But the reasoning behind why it was moved to the parser pkg is because the func returns a map with type devfile type `v1.EndpointExposure`. We wanted to keep the generator pkg for only Kubernetes type functions.

Any thoughts?

maysunfaisal:
@adisky @mik-dass I've added more generator funcs to the list like - GetDeployment, GetService, GetIngress, etc... odo currently calls and uses GetDeploymentSpec, GetServiceSpec, GetIngressSpec for its usage and wraps it around with Deployment, Service, Ingress, etc. to create them. I think if we have these funcs it will help odo have less redundant code. Let me know, what you think.

pkg/devfile/parser/data/v2/components.go
adisky: exposed function should return kube objects, but if it is not intended to be used by consumer, we can move this into generator/utils

maysunfaisal:
@adisky @mik-dass Can this be LGTM'd?

pkg/devfile/parser/data/v2/components.go
maysunfaisal: Updated

pkg/devfile/generator/generators.go
```diff
@@ -0,0 +1,440 @@
...
+// GetDeployment gets a deployment object
+func GetDeployment(deployParams DeploymentParams) *appsv1.Deployment {
+
+	deployment := &appsv1.Deployment{
+		TypeMeta:   deployParams.TypeMeta,
+		ObjectMeta: deployParams.ObjectMeta,
+		Spec:       deployParams.DeploymentSpec,
+	}
+
+	return deployment
```
adisky, line 148: this is getting confusing, can we have a single function instead?
```
GetDeployment
GetDeploymentSpec
```

pkg/devfile/generator/generators.go
```diff
@@ -0,0 +1,440 @@
...
+
+// ServiceParams is a struct that contains the required data to create a service object
+type ServiceParams struct {
+	TypeMeta    metav1.TypeMeta
+	ObjectMeta  metav1.ObjectMeta
+	ServiceSpec corev1.ServiceSpec
+}
+
+// GetService gets the service
+func GetService(serviceParams ServiceParams) *corev1.Service {
```
adisky, line 209: same for
```
GetService
GetServiceSpec
```

adisky (commented):
this way users need to make two calls 

`GetDeploymentSpec()`

and then 

`GetDeployment()`

for simplicity and consistency i think it is better to have one call and all exposed function names be like
GetDeploymentSpec()
GetPodSpec()
GetServiceSpec()
GetIngressSpec()

mik-dass:
> @adisky @mik-dass Can this be LGTM'd?

Please have a look at @adisky 's concerns. Rest looks OK to me.

pkg/devfile/generator/generators.go
maysunfaisal: Currently odo uses `GetDeploymentSpec` and wraps the spec with `deployment` when odo goes to create it. There is 3-4 lines redundant code in odo for every kube resources. How do you feel about `GetDeployment()` being the exposed func and it takes inputs for all the spec generation?

This would mean, odo would need to be modified a bit more.

maysunfaisal: For example, like this https://github.com/openshift/odo/blob/master/pkg/kclient/deployments.go#L169

maysunfaisal: Updated to have GetDeployment, GetService, GetIngress, etc as exposed func

yangcao77 (approved):
Adding lgtm label, since it's already approved by requested reviewers

PR #45, opened 2020-11-12
update the yaml pkg and remove odo specific functions

yangcao77:

### What does this PR do?
this pr update the writer to use `sigs.k8s.io/yaml` as suggested by https://github.com/devfile/api/pull/203#issuecomment-724981014

manual test with odo converter


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/devfile/library
 
-go 1.14
+go 1.13
```
maysunfaisal, line 3: Any reason why we're downgrading to 1.13 here and the other mods?

devfile_test.yaml
```diff
@@ -0,0 +1,42 @@
+commands:
```
maysunfaisal, line 1: is this file being used anywhere

yangcao77: this was added by accident

yangcao77: This was added by accident

go.mod
yangcao77: Since api/odo/console/workspace all use go 1.13.  I want to ensure our library is also using the same go version,to be consistent with the api library and our consumers.

maysunfaisal (approved):
/lgtm

PR #46, opened 2020-12-01
Filtering devfile obj via attributes

maysunfaisal:

### What does this PR do?
Provides an option to filter devfile object via attributes

For example, this helps Console filter components that its interested in rather than parsing all the components belonging to a devfile.yaml

Fixes https://github.com/devfile/api/issues/232

TBD


go.mod
```diff
@@ -16,12 +16,12 @@ require (
...
 	github.com/stretchr/testify v1.6.1 // indirect
 	github.com/xeipuuv/gojsonschema v1.2.0
-	golang.org/x/net v0.0.0-20200324143707-d3edc9973b7e // indirect
 	golang.org/x/sys v0.0.0-20200509044756-6aff5f38e54f // indirect
+	gonum.org/v1/netlib v0.0.0-20190331212654-76723241ea4e // indirect
 	gopkg.in/yaml.v2 v2.3.0 // indirect
-	k8s.io/api v0.18.2
-	k8s.io/apimachinery v0.18.2
+	k8s.io/api v0.18.6
+	k8s.io/apimachinery v0.18.6
```
yangcao77, line 23: can we downgrade `k8s.io/api` and `k8s.io/apimachinery` back to `v0.18.2`? to be consist with api/odo/console/devworkspace?

main.go
```diff
@@ -17,20 +18,39 @@ func main() {
...
-			fmt.Println(d.SchemaVersion)
+			fmt.Printf("schema version: %s\n", d.SchemaVersion)
 		}
 
-		for _, component := range devfile.Data.GetComponents() {
+		compOptions := common.DevfileOptions{
+			Filter: map[string]interface{}{
+				"first": "a",
+				"last":  "c",
+				// "middle": "b",
```
yangcao77, line 28: comments

devfile.yaml
```diff
@@ -2,15 +2,46 @@ schemaVersion: 2.0.0
...
+    alpha.deployment-manifest: https://raw.githubusercontent.com/odo-devfiles/registry/master/devfiles/nodejs/deploy/deployment-manifest.yaml
 starterProjects:
 - name: nodejs-starter
   git:
     remotes:
       origin: https://github.com/odo-devfiles/nodejs-ex.git
 components:
 - name: runtime
+  attributes:
+    first: a
```
yangcao77, line 16: we have more meaningful name if we want it to be a test devfile in the public repo

pkg/devfile/generator/utils.go
```diff
@@ -197,7 +198,7 @@ func getServiceSpec(devfileObj parser.DevfileObj, selectorLabels map[string]stri
 
 	var containerPorts []corev1.ContainerPort
 	portExposureMap := getPortExposure(devfileObj)
-	containers, err := GetContainers(devfileObj)
+	containers, err := GetContainers(devfileObj, common.DevfileOptions{})
```
yangcao77, line 201: should we add `DevfileOptions` to getService as well? to filter out endpoints from specific components / with specific attributes for service creation

go.mod
maysunfaisal: I am not controlling the version, this is automatically updated because devfile/api is using `v0.18.6`

main.go
maysunfaisal: updated

devfile.yaml
maysunfaisal: updated

pkg/devfile/generator/utils.go
maysunfaisal: updated to incl options to service but endpoint will be discussed and added later if necessary since its a level 2 attribute

pkg/devfile/generator/utils_test.go
```diff
@@ -926,7 +1040,7 @@ func TestGetPortExposure(t *testing.T) {
 					Components: tt.containerComponents,
 				},
 			}
-			mapCreated := getPortExposure(devObj)
+			mapCreated, _ := getPortExposure(devObj, tt.filterOptions)
```
yangcao77, line 1043: the returned error should be checked

pkg/devfile/parser/data/v2/commands_test.go
```diff
@@ -191,13 +278,19 @@ func TestDevfile200_UpdateCommands(t *testing.T) {
 
 			d.UpdateCommand(tt.newCommand)
 
-			commandsMap := d.GetCommands()
+			commands, _ := d.GetCommands(common.DevfileOptions{})
```
yangcao77, line 281: same as above, the error should not be ignored. should check unexpected error

pkg/devfile/parser/data/v2/components_test.go
```diff
@@ -213,7 +249,7 @@ func TestGetDevfileContainerComponents(t *testing.T) {
 				},
 			}
 
-			devfileComponents := d.GetDevfileContainerComponents()
+			devfileComponents, _ := d.GetDevfileContainerComponents(tt.filterOptions)
```
yangcao77, line 252: same as above, the error should not be ignored. should check unexpected error

pkg/devfile/generator/utils_test.go
maysunfaisal: true it should be checked. The reason i didnt because err will only be returned when using attributes and i found a bug with attributes in the api repo and doesnt return err properly but i will update this

pkg/devfile/parser/data/v2/commands_test.go
maysunfaisal: same as above

pkg/devfile/parser/data/v2/components_test.go
maysunfaisal: same as above

pkg/devfile/generator/utils_test.go
maysunfaisal: updated

pkg/devfile/parser/data/v2/commands_test.go
maysunfaisal: updated

pkg/devfile/parser/data/v2/components_test.go
maysunfaisal: updated

yangcao77 (approved):
LGTM

PR #48, opened 2020-12-11
use api override and merging utils to get flatterned devfile

yangcao77:
### What does this PR do?
use api override and merging utils to get flatterned devfile
this pr also adds plugin components parsing

https://github.com/devfile/api/issues/198
https://github.com/devfile/api/issues/240

unit test created


yangcao77:
blocked by: https://github.com/devfile/api/issues/276

pkg/devfile/parser/parse.go
```diff
@@ -61,7 +60,20 @@ func Parse(path string) (d DevfileObj, err error) {
 	if err != nil {
 		return d, err
 	}
-	return parseDevfile(d)
+	return parseDevfile(d, true)
+}
+
+// ParseRawDevfile populates the raw devfile data witout overring and merging
```
maysunfaisal, line 66: without* & overriding*
probably also mention in more detail like no parent or plugin will be parsed

pkg/devfile/parser/parse.go
```diff
@@ -90,83 +102,66 @@ func ParseFromData(data []byte) (d DevfileObj, err error) {
...
 
-	// override the parent's components, commands, projects and events
-	err = parentData.OverrideComponents(d.Data.GetParent().Components)
-	if err != nil {
-		return err
-	}
+			parentData, err := ParseFromURL(parent.Uri)
+			if err != nil {
+				return err
+			}
```
maysunfaisal, line 117: We should not check for only URI. We should accept all. But since we dont have the implementation for it, we can have a switch case and err out for the other cases, saying its not supported yet. I think thats neater.

maysunfaisal: Lets also rename the var from `parentData` to `parentDevfileObj` or something. It's very confusing in odo and our library repo to read `parentData.Data.GetDevfileWorkspace()` for example

pkg/devfile/parser/parse.go
```diff
@@ -90,83 +102,66 @@ func ParseFromData(data []byte) (d DevfileObj, err error) {
...
-		return errors.Wrapf(err, "error while adding projects from the parent devfiles")
-	}
-
-	parentStarterProjects, err := parentData.Data.GetStarterProjects(common.DevfileOptions{})
-	if err != nil {
-		return errors.Wrapf(err, "error while getting starter projects from the parent devfiles")
+	for _, component := range components {
+		if component.Plugin != nil && !reflect.DeepEqual(component.Plugin, &v1.PluginComponent{}) {
+			plugin := component.Plugin
+			var pluginData DevfileObj
```
maysunfaisal, line 140: rename here too pls

pkg/devfile/parser/parse.go
```diff
@@ -90,83 +102,66 @@ func ParseFromData(data []byte) (d DevfileObj, err error) {
...
+			if plugin.Uri != "" {
```
maysunfaisal, line 141: Same here as parent, switch case for other types ie; id and kubernetes

pkg/devfile/parser/parse.go
```diff
@@ -90,83 +102,66 @@ func ParseFromData(data []byte) (d DevfileObj, err error) {
...
+				pluginData, err = ParseFromURL(plugin.Uri)
+				if err != nil {
+					return err
+				}
+			}
+			pluginWorkspaceContent := pluginData.Data.GetDevfileWorkspace()
+			result := pluginWorkspaceContent
```
maysunfaisal, line 148: not necessary, I like your flattenedParent var, can we use it here as well? Easier to read.

pkg/devfile/parser/parse.go
```diff
@@ -90,83 +102,66 @@ func ParseFromData(data []byte) (d DevfileObj, err error) {
...
+	if d.Data.GetParent() != nil {
+		if !reflect.DeepEqual(d.Data.GetParent(), &v1.Parent{}) && d.Data.GetParent().Uri != "" {
+			parent := d.Data.GetParent()
 
-	// override the parent's components, commands, projects and events
-	err = parentData.OverrideComponents(d.Data.GetParent().Components)
-	if err != nil {
-		return err
-	}
+			parentData, err := ParseFromURL(parent.Uri)
```
maysunfaisal, line 114: One thing to think about is, you are calling func `ParseFromURL` with both parent and plugin. We need to avoid infinite loops if there the parent or plugin references back to the main devfile. Do we allow such situations acc to the schema?

because it is `parseDevfile(d, true)` inside func `ParseFromURL`

maysunfaisal (changes requested):
Thanks for the changes.

First pass of review..

pkg/devfile/parser/parse.go
maysunfaisal: Not sure if we even need this?

pkg/devfile/parser/data/v2/common/options.go
```diff
@@ -4,6 +4,7 @@ import (
 	"reflect"
 
 	"github.com/devfile/api/pkg/attributes"
+	attriutespkg "github.com/devfile/api/pkg/attributes"
```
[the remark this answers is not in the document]
maysunfaisal: there is also a dup import above

pkg/devfile/parser/parse.go
yangcao77: We will need this, to update an existing devfile using writer will need to write to raw devfile content.

pkg/devfile/parser/parse.go
yangcao77: it would error out since reference back would result in duplication in flattened devfile?

maysunfaisal (commented):
can we add a couple more test cases?

try a test case where there are 
- multiple events of diff type in main, parent and plugin devfile - preStart1 preStart2 in main, preStart1, postStart1 in parent, postStart1, postSart2 in plugin, etc. and see if it holds up nicely. You can probably update the currrent test for events
- different type of overrides - for example, if the plugin devfile has a container component, what happens if the plugin overrides of the same name is of type volume component
- can we also add a test case for this https://github.com/devfile/library/pull/48#discussion_r544552749

maysunfaisal (approved):
/lgtm
thanks!

PR #49, opened 2020-12-14
Add parser tests part 1

mmulholla:
### What does this PR do?
Adds tests which use the parser api to verify, create. and edit valid devfile.yaml files.

Current support : 
Commands: Exec and Composite
Components: Container and Volume

https://github.com/devfile/api/issues/215

Yes. Readme includes instruction for how to modify tests


maysunfaisal:
@mmulholla make a temp commit and print `git status -s` in the CI workflow to see what files are in the git status thats causing it to fail

mmulholla:
I added a change to the build order to enable the build to pass. go fmt was failing because the pull request was running go tests and then checking go fmt. The go test creates a tmp directory which it leaves populated, then go fmt throws an error because of the tmp directory existing. I changed the build order so it checks format and then runs the tests and the build then passes.

I also tweeked build output after running go fmt to output the error message from go fmt - without thus i would never have found this issue!!

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+The tests use the go language and are intended to test every apsect of the parser for every schema attribute. Some basic aspects of the tests:
+
+* A first test (parser_v200_schema_test.go) feeds pre-created devfiles to the parser to ensure the parser can parse all attribues and return an approproate error when the devfile contains an error.
+* A second set of tests (parser_v200_verify_test.go) create devfile content at runtime:
+    * Devfile content is randomly generated and as a result the tests are designed to run multiple times.
+    * Parser functions covered:
+        * Read an existing devfile.
+        * Write a new devfile.
+        * Modify Content of a devfile.
```
yangcao77, line 13:
```suggestion
        * Modify content of a devfile.
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+The tests using pre-created devfiles are complete (but update in progress due to schema changes)
+
+The tests which generate devfiles with random content at run time currently cover the following properties and items.
+
+* Commands: 
+    * Exec
+    * Composite
+* Components 
+    * Container
+    * Volume
```
yangcao77, line 30: since https://github.com/devfile/library/pull/48 has been merged, plugin component with uri defined  is also supported

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Running the tests
+
+1. Go to directory tests/src/tests
+1. Run ```go test``` or ```go test -v```
+1. The test creates the following files: 
+    1. ```./tmp/test.log``` contains log output from the tests.
```
yangcao77, line 37: the `tmp` folder should be added in gitignore

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+    const numThreads = 5		// Number of threads used by multi-thread tests
+    const maxCommands = 10 		// The maximum number of commands to include in a generated devfile
+    const maxComponents = 10	// The maximum number of components to include in a generated devfile
+
+## Basic principles of the tests which randomly generates devfiles
+
+* Each devfile is created in a schema structure.
+* Which attributes are set and the values used are randomized.
+* Once the schema structure is complete it is written in one of two ways.
```
yangcao77, line 90: why? explain in which situation each is being picked to used?

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+        * Exec
+        * Appy (to be implemented)
+        * Composite
+        * VSCodeLaunch (to be implemented)
+        * VSCodeTask (to be implemented)
+    * Each of these command-types have equivalent functions:    
+        * ```func create<command-type>Command() *schema.<command-type>```
+            * creates the command object and calls ```set<command-type>CommandValues``` to add attribute values
+            * for example see: ```func createExecCommand(execCommand *schema.ExecCommand)```
+        * ```func set<command-type>CommandValues(project-sourceProject *schema.<project-source>)```
```
yangcao77, line 163: suggest to switch order of `set<command-type>CommandValues` and `create<command-type>Command()`
should explain it before reference it

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+            * sets random attributes into the provided object 
+            * for example see: ```func setExecCommandValues(execCommand *schema.ExecCommand)```
+    * Functions general to all commands  
+        * ```func (devfile *TestDevfile) addCommand(commandType schema.CommandType) string```
+            * main entry point for a test to add a command
+            * maintains the array of commands in the schema structure
+            * calls generateCommand() 
+        * ```func generateCommand(command *schema.Command, genericCommand *GenericCommand)```
```
yangcao77, line 171: same here, suggest to explain `generateCommand` before reference it under `addCommand`

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+            * includes logic to call the ```create<Command-Type>Command``` function for the command-Type of the supplied command object.
+        * ```func (devfile *TestDevfile) UpdateCommand(command *schema.Command) error```
+            * includes logic to call set<commad-type>CommandValues for each commandType.
+        * ```func (devfile TestDevfile) VerifyCommands(parserCommands []schema.Command) error```
+            * Includes logic to compare the array of commands obtained from the parser with those created by the test. if the compare fails:
```
yangcao77, line 176: be consistent
```suggestion
            * includes logic to compare the array of commands obtained from the parser with those created by the test. If the compare fails:
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+                * each individual command is compared.
+                    * if a command compare fails, the parser version and test version of the command are oputput as yaml files  to the tmp directory 
+                * a check is made to determine if the parser returned a command not known to the test or the pasrer omitted a command expected by the test.
+1. ```test-utils.go```
+    * ```func (devfile TestDevfile) Verify()``` 
+        * Includes code to get object from the paser and verify their content.
+        * For commands code is required to: 
+            1. Retrieve each command from the parser
+            1. Use command id to obtain the GenericCommand object which matches
```
yangcao77, line 185: since defined as `Id` in schema
```suggestion
            1. Use command Id to obtain the GenericCommand object which matches
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+            1. compare the command structure returned by the parser with the command structure saved in the GenericCommand object.
```
yangcao77, line 186: To be consistent
```suggestion
            1. Compare the command structure returned by the parser with the command structure saved in the GenericCommand object.
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+    * ```func (devfile TestDevfile) EditCommands() error```
+        * Specific to command objects.
+            1. Ensure devfile is written to disk
+            1. use parser to read devfile and get all command object
```
yangcao77, line 190:
```suggestion
            1. Use parser to read devfile and get all command object
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+            1.  for each command call:
+                *  ```func (devfile *TestDevfile) UpdateCommand(command *schema.Command) error``` 
+            1. when all commands have been updated, use parser to write the updated devfile to disk
```
yangcao77, line 193:
```suggestion
            1. When all commands have been updated, use parser to write the updated devfile to disk
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+        * Includes code to get object from the paser and verify their content.
+        * For commands code is required to: 
+            1. Retrieve each command from the parser
+            1. Use command id to obtain the GenericCommand object which matches
+            1. compare the command structure returned by the parser with the command structure saved in the GenericCommand object.
+    * ```func (devfile TestDevfile) EditCommands() error```
+        * Specific to command objects.
+            1. Ensure devfile is written to disk
+            1. use parser to read devfile and get all command object
+            1.  for each command call:
```
yangcao77, line 191:
```suggestion
            1.  For each command call:
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+        * includes an array of command types: ```CommandTypes     []schema.CommandType``` 
+    * ```func Test_ExecCommand(t *testing.T)```
+        1. Creates a TestContent object
+        1. Adds a single entry array containg schema.ExecCommandType to the array of command types
+        1. Calls runTest for a single thread test
+        1. Calls runMultiThreadTest for a multi-thread test.
+    * See also
+        *```func Test_<string>ExecCommand(t *testing.T)``` 
+        *```func Test_MultiCommand(t *testing.T)```
+        *```func Test_Everything(t *testing.T)```
```
yangcao77, line 205: Need space after `*` 
```suggestion
        * ```func Test_<string>ExecCommand(t *testing.T)``` 
        * ```func Test_MultiCommand(t *testing.T)```
        * ```func Test_Everything(t *testing.T)```
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+    * add logic to ```func runTest(testContent TestContent, t *testing.T)``` includes logic
```
yangcao77, line 206:
```suggestion
    * Adds logic to ```func runTest(testContent TestContent, t *testing.T)``` which includes logic
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+        1. Add commands to the test
```
yangcao77, line 207:
```suggestion
        1. Adds commands to the test
```

tests/src/tests/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+#### Code flow
+
+Create, modify and verify an exec command: 
+1. parser_v200_verify_test.Test_ExecCommand
+    1. parser-v200-test.runTest
+        1. command-test-utils.AddCommand
+            1. command-test-utils.GenerateCommand
+                1.  command-test-utils.createExecCommand
+                    1. command-test-utils.setExecCommandValues
```
yangcao77, line 219: check the indentation, otherwise the format will be incorrect. 
same same below
```suggestion
        1. command-test-utils.AddCommand
        1. command-test-utils.GenerateCommand
        1.  command-test-utils.createExecCommand
        1. command-test-utils.setExecCommandValues
```

mmulholla:
@yangcao77 Thank you for your comments, I have addressed each of them except:

"since #48 has been merged, plugin component with uri defined is also supported"

the section is about what is supported in the test which does not include that function yet.

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,324 @@
...
+func GetRandomGroupKind() schema.CommandGroupKind {
+	setRandSeed()
+	return GetGroupKinds()[rand.Intn(len(GetGroupKinds()))]
+}
+
+func GetBinaryDecision() bool {
+	return GetRandomDecision(1, 1)
+}
+
+func GetRandomDecision(success int, failure int) bool {
```
yangcao77, line 154: how does this function works? why expect an int value for success and failure?
Besides, I would like to suggest having a simple comment for every function you added

tests/src/tests/command_test_utils.go
```diff
@@ -0,0 +1,256 @@
...
+	return attributes
+}
+
+func addGroup() *schema.CommandGroup {
+
+	commandGroup := schema.CommandGroup{}
+
+	commandGroup.Kind = GetRandomGroupKind()
+	LogMessage(fmt.Sprintf("   ....... group Kind: %s", commandGroup.Kind))
+	commandGroup.IsDefault = GetBinaryDecision()
```
yangcao77, line 40: this might generate an invalid devfile, since each command group can only have one default command. 
check the point 5 in commands validation: https://github.com/devfile/api/blob/master/pkg/validation/validation-rule.md

```
{build, run, test, debug}, each kind of group can only have one default command associated with it. If there are multiple commands of the same kind without a default, a warning will be displayed.
```

mmulholla: At the moment the code only adds one group to any single command, so this is not currently a problem. I may update the code to add multiple in which case I can handle the requirement then.

tests/src/tests/test_utils.go
mmulholla: There are a lots of functions without comments. Is it OK if I add these in a subsequent pull request? The tests are working and would like to get them merged - this is only part 1.

tests/src/tests/command_test_utils.go
yangcao77: Sounds good. might want to open an issue to track for multiple command tests

tests/src/tests/command_test_utils.go
```diff
@@ -0,0 +1,256 @@
...
+			if testCommand, found := getSchemaCommand(devfile.SchemaDevFile.Commands, command.Id); found {
+				if !cmp.Equal(command, *testCommand) {
+					parserFilename := AddSuffixToFileName(devfile.FileName, "_"+command.Id+"_Parser")
+					testFilename := AddSuffixToFileName(devfile.FileName, "_"+command.Id+"_Test")
+					LogMessage(fmt.Sprintf("   .......marshall and write devfile %s", devfile.FileName))
+					c, err := yaml.Marshal(command)
+					if err == nil {
+						err = ioutil.WriteFile(parserFilename, c, 0644)
+					}
+					LogMessage(fmt.Sprintf("   .......marshall and write devfile %s", devfile.FileName))
```
yangcao77, line 228: for the log messages format, e.g. `   .......` and ` --> ERROR:` have to be manually passes in. suggest separate functions:
```
LogErrorMessage(){
testLogger.Println("--> ERROR: " + message)
}
```
and
```
LogInfoMessage(){
testLogger.Println("   ....... " + message)
}
```

mmulholla: Sure, if that is a format you prefer, also its a minor improvement so I prefer to add it in part 2.

tests/src/tests/component_test_utils.go
```diff
@@ -0,0 +1,228 @@
...
+		for _, component := range devfile.SchemaDevFile.Components {
+			components[index] = component
+			index++
+		}
+	} else {
+		components = make([]schema.Component, 1)
+	}
+
+	generateComponent(&components[index], componentType)
+	devfile.SchemaDevFile.Components = components
```
yangcao77, line 54: I am confused with this function.
any reason `generateComponent` requires `&components` instead of just returning a component?
any reason need another copy of `devfile.SchemaDevFile.Components`?

can we just do 
```
devfile.SchemaDevFile.Components = append(devfile.SchemaDevFile.Components, generateComponent( componentType))
```

tests/src/tests/command_test_utils.go
yangcao77: Can you also create an issue to track this as well? I just do not think it's a good idea to manually input the log format for every log message.

tests/src/tests/component_test_utils.go
```diff
@@ -0,0 +1,228 @@
...
+	if GetBinaryDecision() {
+		containerComponent.Env = AddEnv(GetRandomNumber(4))
+	} else {
+		containerComponent.Env = nil
+	}
+
+	if GetBinaryDecision() {
+		containerComponent.VolumeMounts = AddVolume(GetRandomNumber(4))
+	} else {
+		containerComponent.Env = nil
```
yangcao77, line 144: should be `VolumeMounts`?

yangcao77: if it's `nil`, you can just leave it empty instead of setting it as nil. same as above

tests/src/tests/component_test_utils.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return err
+}
+
+func (devfile TestDevfile) VerifyComponents(parserComponents []schema.Component) error {
+
+	LogMessage("Enter VerifyComponents")
+	errorString := ""
+
+	// Compare entire array of commands
```
yangcao77, line 188: components*

tests/src/tests/component_test_utils.go
```diff
@@ -0,0 +1,228 @@
...
+					errorString += LogMessage(fmt.Sprintf(" --> ERROR: Component %s did not match, see files : %s and %s", component.Name, parserFilename, testFilename))
+				} else {
+					LogMessage(fmt.Sprintf(" --> Component matched : %s", component.Name))
+				}
+			} else {
+				errorString += LogMessage(fmt.Sprintf(" --> ERROR: Component from parser not known to test - id : %s ", component.Name))
+			}
+		}
+		for _, component := range devfile.SchemaDevFile.Components {
+			if _, found := getSchemaComponent(parserComponents, component.Name); !found {
```
yangcao77, line 215: since `parserComponents` and `devfile.SchemaDevFile.Components` are already compared using `cmp.Equal`, this for loop is redundant

yangcao77: rechecked,  the loop is executed if the two slices do not match. my bad

tests/src/tests/endpoint-test-utils.go
```diff
@@ -0,0 +1,59 @@
...
+	schema "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
+)
+
+var Exposures = [...]schema.EndpointExposure{schema.PublicEndpointExposure, schema.InternalEndpointExposure, schema.NoneEndpointExposure}
+
+func GetRandomExposure() schema.EndpointExposure {
+	return Exposures[GetRandomNumber(len(Exposures))-1]
+}
+
+var Protocols = [...]schema.EndpointProtocol{schema.HTTPEndpointProtocol, schema.WSEndpointProtocol, schema.TCPEndpointProtocol, schema.UDPEndpointProtocol}
```
yangcao77, line 15: should `wss` and `https` be included as well?

tests/src/tests/command_test_utils.go
```diff
@@ -0,0 +1,256 @@
...
+	if found {
+		LogMessage(fmt.Sprintf(" ....... Updating command id: %s", parserCommand.Id))
+		if testCommand.Exec != nil {
+			setExecCommandValues(parserCommand.Exec)
+		} else if testCommand.Composite != nil {
+			setCompositeCommandValues(parserCommand.Composite)
+		}
+		devfile.replaceSchemaCommand(*parserCommand)
+	} else {
+		errorString += LogMessage(fmt.Sprintf(" ....... Command not found in test : %s", parserCommand.Id))
```
yangcao77, line 93: I would prefer to have `errorString` with type `[]string`
and do `errorString= append(errorString, errorMessage)` instead of using `+=`

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,324 @@
...
+	randomString := string(b)
+	if lower {
+		randomString = strings.ToLower(randomString)
+	}
+	return randomString
+}
+
+func GetGroupKinds() []schema.CommandGroupKind {
+	setRandSeed()
+	return []schema.CommandGroupKind{schema.BuildCommandGroupKind, schema.RunCommandGroupKind, schema.TestCommandGroupKind, schema.DebugCommandGroupKind}
```
yangcao77, line 142: this is a constant slice, no need to setRandSeed. and no need a separate function for the group kinds
suggest to have the slice defined at top of the file, as you've done for protocol and exposure in endpoints

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,324 @@
...
+func GetRandomNumber(max int) int {
+	setRandSeed()
+	return rand.Intn(max) + 1
+}
+
+func GetDevfile(fileName string) TestDevfile {
+	testDevfile := TestDevfile{}
+	testDevfile.SchemaDevFile = schema.Devfile{}
+	testDevfile.FileName = fileName
+	testDevfile.SchemaDevFile.SchemaVersion = "2.0.0"
```
yangcao77, line 168: Just want to point out that I have seen this schema version is hard coded to `2.0.0`, and the tests are relying on devfile structs defined in api repo(schema). However, the parser repo is going to support multiple schema versions, not only V2, but also future versions. 
I know this is the first PR (part1) for the test, we might also want to start considering how the test structure is going to be for future schema versions, i.e. which parts can be commonly used, and which parts are version specific.

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,324 @@
...
+			if err != nil {
+				LogMessage(fmt.Sprintf(" ..... ERROR: setting devfile path : %v", err))
+			} else {
+				devObj := parser.DevfileObj{
+					Ctx:  ctx,
+					Data: newDevfile,
+				}
+				err = devObj.WriteYamlDevfile()
+				if err != nil {
+					LogMessage(fmt.Sprintf(" ..... ERROR: wriring devfile : %v", err))
```
yangcao77, line 208: again, I saw the formats for log messages are inconsistent. 
here is ` ..... ERROR: ` for error messages, and above are using ` --> ERROR`
and I saw some messages are with 5 dots, and some are with 6 dots, and with/without spaces in between. 

this is due to the log format is manually input for every message, and which is something we want to avoid

tests/src/tests/component_test_utils.go
mmulholla: Does append increment the size of a slice?

tests/src/tests/component_test_utils.go
mmulholla: yes, oops

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,324 @@
...
+
+	err := devfile.ParseSchema()
+
+	if err == nil {
+		LogMessage(fmt.Sprintf(" -> Get commands %s : ", devfile.FileName))
+		commands, _ := devfile.ParsedSchemaObj.Data.GetCommands(common.DevfileOptions{})
+		if commands != nil && len(commands) > 0 {
+			err = devfile.VerifyCommands(commands)
+		} else {
+			LogMessage(fmt.Sprintf("  No command found in %s : ", devfile.FileName))
```
yangcao77, line 253:
```suggestion
			LogMessage(fmt.Sprintf("  No commands found in %s : ", devfile.FileName))
```

tests/src/tests/component_test_utils.go
mmulholla: oops

tests/src/tests/component_test_utils.go
yangcao77: different from java array, you do not need to defined size for a slice in golang
you do not need to do `components = make([]schema.Component, currentSize+1)`, as long as you defined the slice as `var components []schema.Component`, you are free to use it.

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,324 @@
...
+				newDevfile.AddCommands(command)
+			}
+			// add components to the new devfile
+			newDevfile.AddComponents(devfile.SchemaDevFile.Components)
+
+			ctx := devfileCtx.NewDevfileCtx(fileName)
+
+			err = ctx.SetAbsPath()
+			if err != nil {
+				LogMessage(fmt.Sprintf(" ..... ERROR: setting devfile path : %v", err))
```
yangcao77, line 200: should return `err` if it's not nil. 
same for other error checks

tests/src/tests/component_test_utils.go
mmulholla: The cmp.Equal doesn't give an error if it fails, so you don't which component(s) was in error. The second loop only runs if the cmp fails and it does a component by component compare so it can narrow down the error to a specific component. Just helps with diagnosis if an error should happen.

tests/src/tests/endpoint-test-utils.go
mmulholla: At the time of writing the tests the parser was not using the schema with this update. If it is now I can extend the test.

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,324 @@
...
+					LogMessage(fmt.Sprintf(" ..... ERROR: wriring devfile : %v", err))
+				}
+			}
+
+		}
+	} else {
+		LogMessage(fmt.Sprintf("   .......marshall and write devfile %s", devfile.FileName))
+		c, err := yaml.Marshal(&(devfile.SchemaDevFile))
+
+		if err == nil {
```
yangcao77, line 217: IMO, we should check `err!= nil`, and return err. 
same for other error checks

tests/src/tests/command_test_utils.go
mmulholla: Sure append is new to me :-), but another I would like to keep for part 2.

tests/src/tests/test_utils.go
mmulholla: Yes, I created endpoint later than components - maybe my go skills were increasing :-)

tests/src/tests/test_utils.go
mmulholla: Yes, that is on my to do list, not sure yet how to best do it.

tests/src/tests/test_utils.go
mmulholla: Yes, that will help with consistency, part 2? :-)

tests/src/tests/endpoint-test-utils.go
yangcao77: yes, it's been merged in api repo as a part of https://github.com/devfile/api/pull/268

tests/src/tests/test_utils.go
mmulholla: The code in this case does return err.

tests/src/tests/test_utils.go
mmulholla: Same here, it does get returned but only at the end. - single entry single exit

tests/src/tests/test_utils.go
mmulholla: lol, the patent attorney in me says that "command" is more accurate.

tests/src/tests/test_utils.go
mmulholla: Have now added simple comments to all functions

tests/src/tests/component_test_utils.go
mmulholla: Done

tests/src/tests/command_test_utils.go
mmulholla: done

tests/src/tests/test_utils.go
mmulholla: Done

tests/src/tests/test_utils.go
mmulholla: Done

tests/src/tests/test_utils.go
mmulholla: Changed all err ==nil to err != nil to hopefully look less weird, but I keep a single exit point.

mmulholla:
@yangcao77 I have hopefully addressed all of your comments to your satisfaction. Also thanks again for the thorough review and teaching me about append.

maysunfaisal (changes requested):
Some review comments

Can we also update the path `tests/src/tests`?

- Dockerfiles can be in `tests/examples/dockerfiles`
- tests code themselves can in `tests/<test type like integration/e2e>` & `tests/utils`
- Readme can in `tests/`

tests/src/tests/command_test_utils.go
```diff
@@ -0,0 +1,256 @@
...
+	var commands []schema.Command
+	index := 0
+	if devfile.SchemaDevFile.Commands != nil {
+		// Commands already exist so expand the current slice
+		currentSize := len(devfile.SchemaDevFile.Commands)
+		commands = make([]schema.Command, currentSize+1)
+		for _, command := range devfile.SchemaDevFile.Commands {
+			commands[index] = command
+			index++
+		}
```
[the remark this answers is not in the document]
maysunfaisal: lol I see you already addressed the append in the latest commit and i missed it by 15 mins 😆

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,394 @@
...
+				}
+			}
+
+		}
+	} else {
+		LogInfoMessage(fmt.Sprintf("Marshall and write devfile %s", devfile.FileName))
+		c, err := yaml.Marshal(&(devfile.SchemaDevFile))
+
+		if err != nil {
+			LogErrorMessage(fmt.Sprintf("Marshall devfile %s : %v", devfile.FileName, err))
```
maysunfaisal, line 269: need to return here because this `err` variable is different from the `var err error` in the first line of the func. So in the end where you `return err` it will be empty when this if block is executed. Same for the if block below this

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,394 @@
...
+	fileName := devfile.FileName
+	if !strings.HasSuffix(fileName, ".yaml") {
+		fileName += ".yaml"
+	}
+
+	if useParser {
+		LogInfoMessage(fmt.Sprintf("Use Parser to write devfile %s", fileName))
+		newDevfile, err := devfileData.NewDevfileData(devfile.SchemaDevFile.SchemaVersion)
+		if err != nil {
+			LogErrorMessage(fmt.Sprintf("Creating new devfile : %v", err))
```
maysunfaisal, line 234: need to return here because this `err` variable is different from the `var err error` in the first line of the func. So in the end where you `return err` it will be empty when this if block is executed. Same for the if block below this

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,394 @@
...
+var infoPrefix = "INFO :"
+
+func LogInfoMessage(message string) string {
+	var infoMessage []string
+	infoMessage = append(infoMessage, infoPrefix, message)
+	return LogMessage(fmt.Sprint(infoMessage))
+}
+
+type TestDevfile struct {
+	SchemaDevFile   schema.Devfile
```
maysunfaisal, line 143: can we rename this to something else? Its very confusing to read `TestDevfile.SchemaDevFile`. Maybe Data or Contents sounds ok?

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,394 @@
...
+				}
+				err = devObj.WriteYamlDevfile()
+				if err != nil {
+					LogErrorMessage(fmt.Sprintf("Writing devfile : %v", err))
+				} else {
+					devfile.SchemaParsed = false
+				}
+			}
+
+		}
```
maysunfaisal, line 263: I see a lot of redundancy here. We are creating a test devfile with `devfile *TestDevfile` which ultimately is the same as `newDevfile, err := devfileData.NewDevfileData(devfile.SchemaDevFile.SchemaVersion)` 

Why not create your test devfile with `newDevfile, err := devfileData.NewDevfileData(devfile.SchemaDevFile.SchemaVersion)` in the first place?

You are storing the test data in two different structs here at this moment

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,394 @@
...
+
+	var err error
+	if !devfile.SchemaParsed {
+		LogInfoMessage(fmt.Sprintf("Parse and Validate %s : ", devfile.FileName))
+		devfile.ParsedSchemaObj, err = devfilepkg.ParseAndValidate(devfile.FileName)
+		if err != nil {
+			LogErrorMessage(fmt.Sprintf("From ParseAndValidate %v : ", err))
+		}
+		devfile.SchemaParsed = true
+	}
```
maysunfaisal, line 293: Not sure what the point of parsing is, when you're curating the devfile struct with fake/dummy data in `CreateDevfile`. So the flow basically is populate the devfile struct, write to file and unmarshall it back to structs in a diff variable.. I think this all can be avoided!

tests/src/tests/command_test_utils.go
```diff
@@ -0,0 +1,258 @@
...
+		command.Composite = createCompositeCommand()
+	}
+	return command
+}
+
+// Update the values of a specified command
+func (devfile *TestDevfile) UpdateCommand(parserCommand *schema.Command) error {
+
+	var errorString []string
+	testCommand, found := getSchemaCommand(devfile.SchemaDevFile.Commands, parserCommand.Id)
```
maysunfaisal, line 72: why are we doing this? You already have the command that you passed in to this function - `parserCommand`. The `getSchemaCommand()` basically loops thru all the commands and matches the id and returns the same command back..

tests/src/tests/command_test_utils.go
```diff
@@ -0,0 +1,258 @@
...
+	if found {
+		LogInfoMessage(fmt.Sprintf("Updating command id: %s", parserCommand.Id))
+		if testCommand.Exec != nil {
+			setExecCommandValues(parserCommand.Exec)
+		} else if testCommand.Composite != nil {
+			setCompositeCommandValues(parserCommand.Composite)
+		}
+		devfile.replaceSchemaCommand(*parserCommand)
+	} else {
+		errorString = append(errorString, LogErrorMessage(fmt.Sprintf("Command not found in test : %s", parserCommand.Id)))
```
maysunfaisal, line 82: Not sure what the point of having an `errorString []string` is here. We are not looping to store multiple errors.

tests/src/tests/command_test_utils.go
```diff
@@ -0,0 +1,258 @@
...
+
+// Use the specified command to replace the command in the schema structure with the same Id.
+func (devfile TestDevfile) replaceSchemaCommand(command schema.Command) {
+	for i := 0; i < len(devfile.SchemaDevFile.Commands); i++ {
+		if devfile.SchemaDevFile.Commands[i].Id == command.Id {
+			devfile.SchemaDevFile.Commands[i] = command
+			break
+		}
+	}
+}
```
maysunfaisal, line 150: I think if we clean up the test devfile structs like I mentioned earlier you can use the parser's updateCommand() func. This is basically duplicating effort and having two copies of the same thing - one in test and one in parser

maysunfaisal (changes requested):
Think test devfile logic can be cleaned up more. There are a few redundant codes in tests.

tests/src/tests/test_utils.go
maysunfaisal: I honestly think we should do that before merging this PR. Once this merged, we will keep adding tests and before we know it, it will be a lot of updates and  refactoring to accommodate future version schema changes.

tests/src/tests/test_utils.go
maysunfaisal: Is this because you want to verify across two structs in the end?

tests/src/tests/README.md
```diff
@@ -0,0 +1,235 @@
+# Devfile Parser Library Tests
+
+## About
+
+The tests use the go language and are intended to test every apsect of the parser for every schema attribute. Some basic aspects of the tests:
+
+* A first test (parser_v200_schema_test.go) feeds pre-created devfiles to the parser to ensure the parser can parse all attribues and return an approproate error when the devfile contains an error.
```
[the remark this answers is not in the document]
mmulholla: Added a note to say the test was not available yet - needs updating as a result of schema changes

tests/src/tests/test_utils.go
```diff
@@ -0,0 +1,324 @@
...
+	devfileCtx "github.com/devfile/library/pkg/devfile/parser/context"
+	devfileData "github.com/devfile/library/pkg/devfile/parser/data"
+	"github.com/devfile/library/pkg/devfile/parser/data/v2/common"
+	"sigs.k8s.io/yaml"
+)
+
+const tmpDir = "./tmp/"
+const logErrorOnly = false
+const logFileName = "test.log"
+const logToFileOnly = true // If set to false the log output will also be output to the console
```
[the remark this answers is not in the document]
mmulholla: Changed the constant format. 
Customizing the settings is using ENV or similar is for the future.

tests/src/tests/test_utils.go
mmulholla: I fixed this to make sure the error is included in return err.

tests/src/tests/test_utils.go
mmulholla: I honestly cannot think of a more descriptive name - going to punt in this one.

tests/src/tests/test_utils.go
mmulholla: The test is meant to test the library api and as a result it should be using every api available, the code my look cumbersome but that is just how tests are. You raise an interesting point though, there may be parts of the api I am not testing fully and I need to do a review of that - make sure I fully exercise all apis.

tests/src/tests/test_utils.go
mmulholla: It can look like madness sometimes but when it comes to tests "that is the way".  
The concept is i create a schema structure with attributes which is now my reference.
I write that as a yaml two ways: using the yaml tool and the parser.
I then use the parser to read the the file, and then check the content the parser returns to me matches my reference content.  Under the covers the parse may do "stuff" which sometimes makes this redundant but  "that is the way" and at least the "stuff" is tested. 
Hope you like the Mandalorian humour :-).

tests/src/tests/command_test_utils.go
mmulholla: Yes, just the way the test work, maintain a reference of devfile content to compare with what the parser returns.

mmulholla:
Guys, thank you for the thorough reviews. I have hopefully responded to all of your comments and addressed most of them.
Please now approve. This is a first pass at adding some parser api tests, at this stage they don't need to be perfect (if tests ever need to be perfect) and they are a work in progress. I know there is a bunch of stuff that is not complete, there are better ways of doing some things, and they are still just touching a tiny percentage of the api, but it is a starting point and an important one. The tests also work.

tests/src/tests/command_test_utils.go
mmulholla: So here I am testing the paser api used to modify existing devfile conent. So first I create the schema structure as reference devfile content. Write that to disk, have the parser read it and then start updating. For updating I update the reference structure, use parser api to make the same update, get the full devfile back and make sure the updates are reflected.  Probably need to get some invalid updates going here to that is for the future.

tests/src/tests/command_test_utils.go
mmulholla: Fixed it.

mmulholla:
@maysunfaisal @yangcao77 Are you OK to approve now?

tests/src/tests/README.md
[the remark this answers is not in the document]
maysunfaisal: If its not there, should we really merge this information? Dont want others to be confused whoever is reading it

tests/src/tests/command_test_utils.go
maysunfaisal: I get what you mean, but you are not doing anything with the `testCommand` in this func:

```
var err error
	testCommand, found := getSchemaCommand(devfile.SchemaDevFile.Commands, parserCommand.Id)
	if found {
		LogInfoMessage(fmt.Sprintf("Updating command id: %s", parserCommand.Id))
		if testCommand.Exec != nil {
			setExecCommandValues(parserCommand.Exec)
		} else if testCommand.Composite != nil {
			setCompositeCommandValues(parserCommand.Composite)
		}
		devfile.replaceSchemaCommand(*parserCommand)
	} else {
		err = errors.New(LogErrorMessage(fmt.Sprintf("Command not found in test : %s", parserCommand.Id)))
	}
	return err
```

tests/utils/test_utils.go
```diff
@@ -52,37 +55,23 @@ func init() {
...
-	_, fn, _, ok := runtime.Caller(1)
-	if !ok {
-		return tmpDir
-	}
-	testFile := filepath.Base(fn)
-	testFileExtension := filepath.Ext(testFile)
-	subdir := testFile[0 : len(testFile)-len(testFileExtension)]
-	return CreateTempDir(subdir)
+	testError(true)
+	testError1(true)
```
maysunfaisal, line 60: Are these func and their calls supposed to be removed?

tests/utils/project_test_utils.go
```diff
@@ -0,0 +1,10 @@
+package utils
+
+import (
+	schema "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
+)
+
+type GenericProject struct {
+	Name    string
+	Project *schema.Project
+}
```
maysunfaisal, line 10: pls delete this file and struct, its not being used anywhere

tests/utils/test_utils.go
```diff
@@ -0,0 +1,433 @@
...
+	var returnError error
+	if len(errorString) > 0 {
+		returnError = errors.New(fmt.Sprint(errorString))
+	}
+	return returnError
+
+}
+
+// EditCommands modifies random attributes for each of the commands in the devfile.
+func (devfile TestDevfile) EditCommands() error {
```
maysunfaisal, line 326: should this be in command utils file?

tests/utils/test_utils.go
```diff
@@ -0,0 +1,433 @@
...
+		}
+		LogInfoMessage(fmt.Sprintf("Write updated file to yaml : %s", devfile.FileName))
+		devfile.ParsedSchemaObj.WriteYamlDevfile()
+		devfile.SchemaParsed = false
+	}
+	return err
+}
+
+// EditComponents modifies random attributes for each of the components in the devfile.
+func (devfile TestDevfile) EditComponents() error {
```
maysunfaisal, line 353: should this be in component utils file?

tests/utils/endpoint-test-utils.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+var Exposures = [...]schema.EndpointExposure{schema.PublicEndpointExposure, schema.InternalEndpointExposure, schema.NoneEndpointExposure}
+
+// GetRandomExposure returns a random exposure value
+func GetRandomExposure() schema.EndpointExposure {
+	return Exposures[GetRandomNumber(len(Exposures))-1]
+}
+
+//var Protocols = [...]schema.EndpointProtocol{schema.HTTPEndpointProtocol, schema.HTTPSEndpointProtocol, schema.WSEndpointProtocol, schema.WSSEndpointProtocol, schema.TCPEndpointProtocol, schema.UDPEndpointProtocol}
```
maysunfaisal, line 16: remove comment

tests/utils/endpoint-test-utils.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+
+	schema "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
+)
+
+var Exposures = [...]schema.EndpointExposure{schema.PublicEndpointExposure, schema.InternalEndpointExposure, schema.NoneEndpointExposure}
+
+// GetRandomExposure returns a random exposure value
+func GetRandomExposure() schema.EndpointExposure {
```
maysunfaisal, line 12: func can be unexported, used only within the pkg

tests/utils/endpoint-test-utils.go
```diff
@@ -0,0 +1,63 @@
...
+	return Exposures[GetRandomNumber(len(Exposures))-1]
+}
+
+//var Protocols = [...]schema.EndpointProtocol{schema.HTTPEndpointProtocol, schema.HTTPSEndpointProtocol, schema.WSEndpointProtocol, schema.WSSEndpointProtocol, schema.TCPEndpointProtocol, schema.UDPEndpointProtocol}
+var Protocols = [...]schema.EndpointProtocol{schema.HTTPEndpointProtocol, schema.WSEndpointProtocol, schema.TCPEndpointProtocol, schema.UDPEndpointProtocol}
+
+// GetRandomProtocol returns a random protocol value
+func GetRandomProtocol() schema.EndpointProtocol {
```
maysunfaisal, line 20: func can be unexported, used only within the pkg

tests/utils/endpoint-test-utils.go
```diff
@@ -0,0 +1,63 @@
...
+	return Protocols[GetRandomNumber(len(Protocols))-1]
+}
+
+// CreateEndpoints createa and returns a randon numebr of endpoints in a schema structure
+func CreateEndpoints() []schema.Endpoint {
```
maysunfaisal, line 25: func can be unexported, used only within the pkg

tests/utils/command_test_utils.go
```diff
@@ -0,0 +1,253 @@
...
+	"io/ioutil"
+
+	"github.com/google/go-cmp/cmp"
+	"sigs.k8s.io/yaml"
+
+	schema "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
+)
+
+// AddEnv creates and returns a specifed number of env attributes in a schema structure
+func AddEnv(numEnv int) []schema.EnvVar {
```
maysunfaisal, line 15: func can be unexported, used only within the pkg

pls check for all funcs in this file.

tests/utils/component_test_utils.go
```diff
@@ -0,0 +1,233 @@
...
+	"io/ioutil"
+	"strconv"
+
+	schema "github.com/devfile/api/pkg/apis/workspaces/v1alpha2"
+	"github.com/google/go-cmp/cmp"
+	"sigs.k8s.io/yaml"
+)
+
+// AddVolume returns volumeMounts in a schema structure based on a specified number of volumes
+func AddVolume(numVols int) []schema.VolumeMount {
```
maysunfaisal, line 15: func can be unexported, used only within the pkg

pls check for all funcs in this file.

tests/src/tests/command_test_utils.go
mmulholla: testCommand is used in the if and else statements. The logic is basically

1. Use the id from the command returned by the parser to find the equivalent command in the test.
2: if it is found:
    a. check the type of the command recorded in the test and set appropriate update values in the command returned by the parser
   b. replace the command stored in the test with the new version to be given to the parser.
3. If it is not found we have a problem.

on return the parser api is use to write the updated parser command.

I could do this any number of ways but the outcome needs to be the same.

tests/utils/test_utils.go
```diff
@@ -0,0 +1,433 @@
...
+			err = devfile.VerifyCommands(commands)
+			if err != nil {
+				errorString = append(errorString, LogErrorMessage(fmt.Sprintf("Verfify Commands %s : %v", devfile.FileName, err)))
+			}
+		} else {
+			LogInfoMessage(fmt.Sprintf("No command found in %s : ", devfile.FileName))
+		}
+	}
+
+	if err == nil {
```
maysunfaisal, line 304: no point is another if block? You can have the logic in the same block

tests/utils/command_test_utils.go
```diff
@@ -0,0 +1,253 @@
...
+			} else {
+				errorString = append(errorString, LogErrorMessage(fmt.Sprintf("Command from parser not known to test - id : %s ", command.Id)))
+			}
+
+		}
+		for _, command := range devfile.SchemaDevFile.Commands {
+			if _, found := getSchemaCommand(parserCommands, command.Id); !found {
+				errorString = append(errorString, LogErrorMessage(fmt.Sprintf("Command from test not returned by parser : %s ", command.Id)))
+			}
+		}
```
maysunfaisal, line 243: Any reason we are looping again over here? We just looped thru test devfile and parsed commands in the above for loop.

tests/utils/component_test_utils.go
```diff
@@ -0,0 +1,233 @@
...
+				}
+			} else {
+				errorString = append(errorString, LogErrorMessage(fmt.Sprintf("Component from parser not known to test - id : %s ", component.Name)))
+			}
+		}
+		for _, component := range devfile.SchemaDevFile.Components {
+			if _, found := getSchemaComponent(parserComponents, component.Name); !found {
+				errorString = append(errorString, LogErrorMessage(fmt.Sprintf("Component from test not returned by parser : %s ", component.Name)))
+			}
+		}
```
maysunfaisal, line 223: Same as the command verify comment.

maysunfaisal (changes requested):
Few other minor nits and comments

tests/utils/test_utils.go
mmulholla: oops.

tests/utils/project_test_utils.go
mmulholla: Will remove the file. I have a branch waiting on this which implements this fully.

tests/utils/test_utils.go
mmulholla: I prefer where it is because the use the function makes of the parser, wanted to keep the command_utils pure command stuff.

tests/utils/test_utils.go
mmulholla: I prefer where it is because the use the function makes of the parser, wanted to keep the component_utils pure component stuff.

tests/utils/endpoint-test-utils.go
mmulholla: It is a place holder for when I can run against the latest schema, at which point it will be uncomented.

tests/utils/component_test_utils.go
mmulholla: Done. Envisioned a time i the future when tests may get this specific and need these apis but unexported for now.

tests/utils/test_utils.go
mmulholla: Moved.

tests/utils/command_test_utils.go
mmulholla: The previous loop goes through all the commands returned by the parser and makes sure they are the same as those expected by test. This loop does a quick check that the test does not have commands that the parser did not return.

tests/utils/component_test_utils.go
mmulholla: The previous loop goes through all the components returned by the parser and makes sure they are the same as those expected by test. This loop does a quick check that the test does not have components that the parser did not return.

mmulholla:
@maysunfaisal - all done now?

maysunfaisal (approved):
/lgtm

PR #54, opened 2021-01-27
Add tests which use parser to verify schema

mmulholla:
Sorry, a lot in this PR.

I updated the directory structure to be cleaner which affected the existing tests which all moved.
I also added the new schema test which basically feeds a bunch of pre-created devfiles. mostly invalid, into the parser to verify the schema - similar to the tests in the api repo.
The updated  README hopefully explains things.  

Here is the branch which might make easier viewing: https://github.com/mmulholla/library/tree/AddSchemaVerifyTestsPart1/tests

There are two new go files which are the main part of the update:
https://github.com/mmulholla/library/blob/AddSchemaVerifyTestsPart1/tests/v2/schemaTest/schema_test.go
https://github.com/mmulholla/library/blob/AddSchemaVerifyTestsPart1/tests/v2/utils/schemaTest_utils.go


mmulholla:
Build failure is unrelated to this PR:

ok  	github.com/devfile/library/pkg/devfile/parser/data	0.009s
--- FAIL: TestGetDevfileContainerComponents (0.00s)
    --- FAIL: TestGetDevfileContainerComponents/Case_4_:_Get_Container_component_with_the_specified_filter (0.00s)
        components_test.go:285: TestGetDevfileContainerComponents unexpected error: Attribute with key "secondString" does not exist
FAIL
FAIL	github.com/devfile/library/pkg/devfile/parser/data/v2	0.006s

The test added by this pass:

ok  	github.com/devfile/library/tests/v2/parserTest	38.706s
ok  	github.com/devfile/library/tests/v2/schemaTest	0.736s

tests/v2/devfiles/components/kubernetesMixedCaseName.yaml
```diff
@@ -0,0 +1,3 @@
+  - name: MixedCaseName
+    kubernetes:
+      inlined: testInlined
```
maysunfaisal, line 3: nit: new line

tests/v2/devfiles/components/pluginAll.yaml
```diff
@@ -0,0 +1,47 @@
+  - name: "testpluginuri"
+    plugin:
+      uri: "https://raw.githubusercontent.com/devfile/library/master/tests/v2/devfiles/samples/testParent.yaml"
```
maysunfaisal, line 3: no kubernetes or id in snippet, but look out for the PR to remove plugin. so we may need to delete these tests

tests/v2/devfiles/metadata/metadataAll.yaml
```diff
@@ -0,0 +1,15 @@
...
+  version: 1.0.0
+metadata:
+  name: test-metadata2
+metadata:  
+  version: 1.0.b 
+metadata:
+  attributes: 
+    test: metadata
+    outcome: pass
+
```
maysunfaisal, line 15: can we have multiple metadata top levels like this? Is this expected to fail?

tests/v2/schemaTest/schema_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Run each test
+	for _, testToRun := range testsToRun {
+
+		totalTests++
+
+		// Test may be disabled pending changes to the schema
+		if testToRun.Disabled {
+			t.Log(utils.LogInfoMessage(fmt.Sprintf("SKIP : %s", testToRun.FileName)))
+			skippedTests++
```
maysunfaisal, line 43: Please mention about the `Disabled` property in README

maysunfaisal (changes requested):

tests/v2/utils/test_utils.go
```diff
@@ -0,0 +1,375 @@
...
+}
+
+// CreateTempDir creates a specified sub directory under the temp directory if it does not exist.
+// Returns the name of the created directory.
+func CreateTempDir(subdir string) string {
+	tempDir := tmpDir + subdir + "/"
+	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
+		os.Mkdir(tempDir, 0755)
+	}
+	return tempDir
```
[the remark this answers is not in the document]
mmulholla: If I can't check or create a tmp directory at this point in the test, bigger problems exist and we probably never get this far, so adding logic for it and the ripple effect it has on callers of the function adds over complication for something which should never happen and it of does worse things will be going wrong. I often say I am writing test for the parser not the operating system.

tests/v2/utils/schemaTest_utils.go
```diff
@@ -0,0 +1,146 @@
...
+				yamlFile.WriteString("\n")
+			}
+		}
+		yamlFile.Close()
+	}
+	return yamlFileName, err
+}
+
+// GetAlTests returns all tests from all json files in the specified directory
+func GetAllTests(directory string) ([]TestToRun, bool) {
```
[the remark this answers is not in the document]
mmulholla: No hiding going on, the test puts out an error message for every error but when the error happens. The thing here is that this function is reading all of the tests from all of the json files, and potentially there could be many many errors, but at the same time some valid test will probably have been identified and then those tests should be run. As the code which reads the return boolean outputs: 

fmt.Sprintf("At least one error occurred reading tests from %s, see log for details", jsonDir))) 

Question should not be what others normally do in go does but what is best for the situation. If you want an err I would create one before return and set it to the message above, but why?

tests/v2/utils/schemaTest_utils.go
```diff
@@ -0,0 +1,146 @@
...
+
+	var err error
+	if len(testJsonFile.Tests) < 1 {
+		// Open the json file which defines the tests to run
+		testJson, err := os.Open(filepath.Join(jsonDir, testJsonFile.FileInfo.Name()))
+		if err != nil {
+			LogErrorMessage(fmt.Sprintf("Failed to open %s : %s", testJsonFile.FileInfo.Name(), err))
+		} else {
+			// Read contents of the json file which defines the tests to run
+			byteValue, err := ioutil.ReadAll(testJson)
```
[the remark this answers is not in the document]
mmulholla: Fixed. I hate go! It is like Java with handcuffs.

tests/v2/utils/schemaTest_utils.go
```diff
@@ -0,0 +1,146 @@
...
+	return jsonFiles, err
+}
+
+// GetTests returns an array of TestToRun objects, one for each test contained in a json file
+func (testJsonFile *TestJsonFile) GetTests() ([]TestToRun, error) {
+
+	var err error
+	if len(testJsonFile.Tests) < 1 {
+		// Open the json file which defines the tests to run
+		testJson, err := os.Open(filepath.Join(jsonDir, testJsonFile.FileInfo.Name()))
```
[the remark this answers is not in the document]
mmulholla: Fixed. I hate go! It is like Java with handcuffs.

tests/v2/devfiles/commands/compositeMinimum.yaml
```diff
@@ -0,0 +1,4 @@
+  - id: compositetest1
+    composite:
+      group:
+        kind: test
```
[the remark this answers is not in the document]
mmulholla: Nothing in composite is mandatory, does seem odd. Could it make sense in a parent or plugin?

tests/v2/devfiles/components/kubernetesMixedCaseName.yaml
mmulholla: done

tests/v2/devfiles/components/pluginAll.yaml
mmulholla: yes, parser currently rejects anything other than uri so I took kubernetes and id out. 
This error from parser: "plugin URI undefined, currently only URI is suppported"

tests/v2/devfiles/metadata/metadataAll.yaml
mmulholla: Have thought about this myself, so far nothing complains about this format so I have gone with it.

tests/v2/schemaTest/schema_test.go
mmulholla: Done.

tests/README.md
```diff
@@ -28,24 +28,77 @@ The tests which generate devfiles with random content at run time currently cove
...
+    - Missing mandatory attributes.
+    - Multiple use of a one-of attribute.
+    - Attribute values of the wrong type.
+
+### Test structure
+
+- ```test/v2/devfiles``` : contains yaml snippets which are used to generate yaml files for the tests. The names of the sub-directories and files should reflect their purpose.
+- ```test/v2/schemaTest/schema-test.go``` : the go unit test program.
+- ```test/v2/json``` :  contains the json files which define the tests which the test program will run:
+    - ```test-xxxxxxx.json``` : these files are the top level json files, they define the schema to verify and the test files to run.
```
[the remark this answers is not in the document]
mmulholla: Fixed

mmulholla:
@maysunfaisal Thank you for the review. I have responded to all comments and fixed most.

PR #57, opened 2021-02-01
relative uri support

yangcao77:
### What does this PR do?
This PR adds relative URI support. 
`parseFromURI` checks the current devfile's abspath/url, and append to get the target uri and parse from that uri
 

https://github.com/devfile/api/issues/309

added unit test to test for relative uris

### DIFF
--- a/pkg/devfile/parser/context/context.go
+++ b/pkg/devfile/parser/context/context.go
@@ -1,12 +1,15 @@
 package parser
 
 import (
+	"fmt"
 	"net/url"
 
 	"github.com/devfile/library/pkg/testingutil/filesystem"
 	"github.com/devfile/library/pkg/util"
 	"k8s.io/klog"
 )
+
+var URIMap = make(map[string]bool)
 
 // DevfileCtx stores context info regarding devfile
 type DevfileCtx struct {
@@ -67,6 +70,10 @@ func (d *DevfileCtx) Populate() (err error) {
 		return err
 	}
 	klog.V(4).Infof("absolute devfile path: '%s'", d.absPath)
+	if _, exist := URIMap[d.absPath]; exist {
+		return fmt.Errorf("URI %v is recursively referenced", d.absPath)
+	}
+	URIMap[d.absPath] = true
 	// Read and save devfile content
 	if err := d.SetDevfileContent(); err != nil {
 		return err
@@ -81,7 +88,10 @@ func (d *DevfileCtx) PopulateFromURL() (err error) {
 	if err != nil {
 		return err
 	}
-
+	if _, exist := URIMap[d.url]; exist {
+		return fmt.Errorf("URI %v is recursively referenced", d.url)
+	}
+	URIMap[d.url] = true
 	// Read and save devfile content
 	if err := d.SetDevfileContent(); err != nil {
 		return err
@@ -101,8 +111,14 @@ func (d *DevfileCtx) Validate() error {
 	return d.ValidateDevfileSchema()
 }
 
+// GetAbsPath func returns current devfile absolute path
 func (d *DevfileCtx) GetAbsPath() string {
 	return d.absPath
+}
+
+// GetURL func returns current devfile absolute URL address
+func (d *DevfileCtx) GetURL() string {
+	return d.url
 }
 
 // SetAbsPath sets absolute file path for devfile
--- a/pkg/devfile/parser/context/context.go
+++ b/pkg/devfile/parser/context/context.go
@@ -70,7 +70,7 @@ func (d *DevfileCtx) Populate() (err error) {
 		return err
 	}
 	klog.V(4).Infof("absolute devfile path: '%s'", d.absPath)
-	if _, exist := URIMap[d.absPath]; exist {
+	if URIMap[d.absPath] {
 		return fmt.Errorf("URI %v is recursively referenced", d.absPath)
 	}
 	URIMap[d.absPath] = true
@@ -88,7 +88,7 @@ func (d *DevfileCtx) PopulateFromURL() (err error) {
 	if err != nil {
 		return err
 	}
-	if _, exist := URIMap[d.url]; exist {
+	if URIMap[d.url] {
 		return fmt.Errorf("URI %v is recursively referenced", d.url)
 	}
 	URIMap[d.url] = true
--- a/pkg/devfile/parser/parse.go
+++ b/pkg/devfile/parser/parse.go
@@ -16,6 +16,7 @@ import (
 
 	v1 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	apiOverride "github.com/devfile/api/v2/pkg/utils/overriding"
+	"github.com/devfile/api/v2/pkg/validation"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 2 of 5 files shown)


pkg/devfile/parser/parse.go
```diff
@@ -181,3 +178,34 @@ func parseParentAndPlugin(d DevfileObj) (err error) {
 
 	return nil
 }
+
+func parseFromURI(uri string, curDevfile DevfileObj) (DevfileObj, error) {
```
maysunfaisal, line 182: looking at the tests, i would just pass the devfile context here rather than the entire devfile object

pkg/devfile/parser/parse_test.go
```diff
@@ -2395,3 +2398,298 @@ func Test_parseParentAndPlugin_RecursivelyReference_withMultipleURI(t *testing.T
...
+			}
+			got, err := parseFromURI(tt.uri, tt.curDevfile)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Test_parseFromURI() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			if err != nil && tt.wantErr {
+				return
+			}
```
maysunfaisal, line 2688: lets try switching to this if condition, it handles everything `if tt.wantErr == (err == nil) {}`

pkg/devfile/parser/parse_test.go
```diff
@@ -2395,3 +2398,298 @@ func Test_parseParentAndPlugin_RecursivelyReference_withMultipleURI(t *testing.T
...
+			},
+			uri:     invalidURL,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// if the main devfile is from local, need to set absolute path
+			if tt.curDevfile.Ctx.GetURL() == "" {
+				err := tt.curDevfile.Ctx.SetAbsPath()
```
maysunfaisal, line 2674: how come you dont need to set it for url since we dont call Populate() for the current devfile 🤔

pkg/devfile/parser/parse_test.go
```diff
@@ -2395,3 +2398,298 @@ func Test_parseParentAndPlugin_RecursivelyReference_withMultipleURI(t *testing.T
...
+			name: "case 4: should fail if file not exist",
+			curDevfile: DevfileObj{
+				Ctx: devfileCtx.NewDevfileCtx(OutputDevfileYamlPath),
+				Data: &v2.DevfileV2{
+					Devfile: v1.Devfile{
+						DevWorkspaceTemplateSpec: v1.DevWorkspaceTemplateSpec{
+							Parent: &v1.Parent{
+								ImportReference: v1.ImportReference{
+									ImportReferenceUnion: v1.ImportReferenceUnion{
+										Uri: "notexist/devfile.yaml",
```
maysunfaisal, line 2613: `notExistURI`

pkg/devfile/parser/parse_test.go
```diff
@@ -2395,3 +2398,298 @@ func Test_parseParentAndPlugin_RecursivelyReference_withMultipleURI(t *testing.T
...
+									},
+								},
+							},
+							DevWorkspaceTemplateSpecContent: v1.DevWorkspaceTemplateSpecContent{},
+						},
+					},
+				},
+			},
+			uri:     "notexist/devfile.yaml",
```
maysunfaisal, line 2622: `notExistURI`

maysunfaisal (commented):
Some questions on tests and minor updates

pkg/devfile/parser/context/context.go
```diff
@@ -67,6 +70,10 @@ func (d *DevfileCtx) Populate() (err error) {
 		return err
 	}
 	klog.V(4).Infof("absolute devfile path: '%s'", d.absPath)
+	if _, exist := URIMap[d.absPath]; exist {
```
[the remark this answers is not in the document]
maysunfaisal: same for the other function

pkg/devfile/parser/parse.go
```diff
@@ -181,3 +178,34 @@ func parseParentAndPlugin(d DevfileObj) (err error) {
...
+	// validate URI
+	err := validation.ValidateURI(uri)
+	if err != nil {
+		return DevfileObj{}, err
+	}
+
+	// absolute URL address
+	if strings.HasPrefix(uri, "http://") || strings.HasPrefix(uri, "https://") {
```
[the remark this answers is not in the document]
yangcao77: these conditions check 3 different values, so I did not use switch case
the first if checks `uri`, the 2nd and 3rd checks ctx.abspath and ctx.url.

pkg/devfile/parser/parse_test.go
yangcao77: because url is an absolute url already. it's being set when creating the contect `NewURLDevfileCtx`
however, when creating context with `NewDevfileCtx`, it only sets the `relPath`; so need to call an additional function `SetAbsPath()` to set the `absPath`

pkg/devfile/parser/context/context.go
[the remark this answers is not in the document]
yangcao77: updated

pkg/devfile/parser/parse.go
```diff
@@ -181,3 +178,34 @@ func parseParentAndPlugin(d DevfileObj) (err error) {
...
+
+	if curDevfile.Ctx.GetURL() != "" {
+		u, err := url.Parse(curDevfile.Ctx.GetURL())
+		if err != nil {
+			return DevfileObj{}, err
+		}
+
+		u.Path = path.Join(path.Dir(u.Path), uri)
+		jointURL := u.String()
+		return ParseFromURL(jointURL)
```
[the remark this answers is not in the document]

pkg/devfile/parser/parse_test.go
```diff
@@ -2395,3 +2398,298 @@ func Test_parseParentAndPlugin_RecursivelyReference_withMultipleURI(t *testing.T
...
+		t.Errorf("fail to write to file: %v", err)
+		return
+	}
+	defer os.RemoveAll("testTmp/")
+
+	parentDevfile := DevfileObj{
+		Ctx: devfileCtx.NewURLDevfileCtx(httpPrefix + uri1),
+		Data: &v2.DevfileV2{
+			Devfile: v1.Devfile{
+				DevWorkspaceTemplateSpec: v1.DevWorkspaceTemplateSpec{
```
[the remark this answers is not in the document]
yangcao77: good catch, updated

pkg/devfile/parser/parse.go
yangcao77: good point! updated

pkg/devfile/parser/parse_test.go
yangcao77: updated

pkg/devfile/parser/parse_test.go
yangcao77: updated

pkg/devfile/parser/parse_test.go
yangcao77: remove this one, as only ctx is required as you pointed out

pkg/devfile/parser/parse_test.go
maysunfaisal: we should probably update `NewDevfileCtx` next time as well so that we dont need an extra call in Populate

maysunfaisal (approved):
lgtm

PR #58, opened 2021-02-04
validation consolidation

yangcao77:
### What does this PR do?
This PR integrates validation into devfile/libaray
also updates the api test to generate valid devfiles for tests

https://github.com/devfile/api/issues/337

### DIFF
--- a/tests/api/parser_v200_verify_test.go
+++ b/tests/api/parser_v200_verify_test.go
@@ -22,10 +22,10 @@ const (
 
 // TestContent - structure used by a test to configure the tests to run
 type TestContent struct {
-	CommandTypes     []schema.CommandType
-	ComponentTypes   []schema.ComponentType
-	FileName         string
-	EditContent      bool
+	CommandTypes   []schema.CommandType
+	ComponentTypes []schema.ComponentType
+	FileName       string
+	EditContent    bool
 }
 
 func Test_ExecCommand(t *testing.T) {
@@ -45,7 +45,6 @@ func Test_ExecCommandEdit(t *testing.T) {
 	runMultiThreadTest(testContent, t)
 }
 
-
 func Test_CompositeCommand(t *testing.T) {
 	testContent := TestContent{}
 	testContent.CommandTypes = []schema.CommandType{schema.CompositeCommandType}
@@ -63,11 +62,10 @@ func Test_CompositeCommandEdit(t *testing.T) {
 	runMultiThreadTest(testContent, t)
 }
 
-
 func Test_MultiCommand(t *testing.T) {
 	testContent := TestContent{}
 	testContent.CommandTypes = []schema.CommandType{schema.ExecCommandType,
-													schema.CompositeCommandType}
+		schema.CompositeCommandType}
 	testContent.EditContent = true
 	testContent.FileName = utils.GetDevFileName()
 	runTest(testContent, t)
@@ -153,9 +151,9 @@ func runMultiThreadTest(testContent TestContent, t *testing.T) {
 func runTest(testContent TestContent, t *testing.T) {
 
 	utils.LogMessage(fmt.Sprintf("Start test for %s", testContent.FileName))
-	testDevfile,err := utils.GetDevfile(testContent.FileName)
-	if err!=nil {
-		t.Fatalf(utils.LogMessage(fmt.Sprintf("Error creating devfile : %v",err)))
+	testDevfile, err := utils.GetDevfile(testContent.FileName)
+	if err != nil {
+		t.Fatalf(utils.LogMessage(fmt.Sprintf("Error creating devfile : %v", err)))
 	}
 
 	if len(testContent.CommandTypes) > 0 {
--- a/tests/api/parser_v200_verify_test.go
+++ b/tests/api/parser_v200_verify_test.go
@@ -25,14 +25,12 @@ type TestContent struct {
 	CommandTypes     []schema.CommandType
 	ComponentTypes   []schema.ComponentType
 	FileName         string
-	CreateWithParser bool
 	EditContent      bool
 }
 
 func Test_ExecCommand(t *testing.T) {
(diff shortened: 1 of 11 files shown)


mmulholla:
Notes on test updates. I had to rewire the tests for validations: 

1. When testing an exec command a component container is also needed with the command id listed. 
2. When testing a component container with a list of commands the commands must also exists in the devfile.
3. When testing component container with a volume to component volume must also exist.
4. When testing a composite command the commands listed must also exist.
5. Groups, each kind added to the devfile requires at least one and no more that one defaults.
6. The ports specified in endpoints must be unique.

Apart from the obvious addition of logic for these I added, for example, commandAdded (called whenever a command is created) and commandUpadated (called whenever a command is updated). These methods keep the test and parser data updated for when they are written and actually simplifies other areas of the code. However this is key to supporting edits that affect, for example, both commands and components.

tests/utils/component_test_utils.go
```diff
@@ -37,62 +51,84 @@ func getSchemaComponent(components []schema.Component, name string) (*schema.Com
...
 	containerComponent.Image = GetRandomUniqueString(8+GetRandomNumber(10), false)
 
 	if GetBinaryDecision() {
 		numCommands := GetRandomNumber(3)
-		containerComponent.Command = make([]string, numCommands)
 		for i := 0; i < numCommands; i++ {
-			containerComponent.Command[i] = GetRandomString(4+GetRandomNumber(10), false)
-			LogInfoMessage(fmt.Sprintf("....... command %d of %d : %s", i, numCommands, containerComponent.Command[i]))
+			commandId := GetRandomString(4+GetRandomNumber(10), true)
+			containerComponent.Command = append(containerComponent.Command, commandId)
```
yangcao77, line 130: The `command` field under container component is unrelated to `Commands`  defined in devfile. 
`command` and `args` in container are actually the docker container entrypoints(in dockerfile)

This restriction should be removed

tests/utils/component_test_utils.go
```diff
@@ -130,42 +166,45 @@ func setContainerComponentValues(containerComponent *schema.ContainerComponent)
...
-		containerComponent.VolumeMounts = nil
+	if len(containerComponent.VolumeMounts) == 0 {
+		if GetBinaryDecision() {
+			containerComponent.VolumeMounts = devfile.addVolume(GetRandomNumber(4))
+		}
 	}
 
 	if GetBinaryDecision() {
-		containerComponent.Endpoints = CreateEndpoints()
+		containerComponent.Endpoints = devfile.CreateEndpoints()
```
yangcao77, line 176: Are we going  to test component with multiple endpoints in the future?
Just to note that, the port defined in endpoints has to be unique across components, but same container can have multiple endpoints sharing  the same port.  

So the restriction looks correct in the current test coverage, but are we looking to expend the coverage?

tests/utils/component_test_utils.go
mmulholla: removed. Main change was when I created an exec command I called a function to add the command to the container - now the function just returns the name of a container without adding the command to it.

tests/utils/component_test_utils.go
mmulholla: Test already includes multiple endpoints. Updated the test so that it will randomly use the same port when adding multiple endpointd to a given container.

tests/utils/command_test_utils.go
```diff
@@ -34,79 +47,81 @@ func addAttributes(numAtrributes int) map[string]string {
...
-
-// generateCommand creates a command of a specified type in a schema structure
-func generateCommand(commandType schema.CommandType) schema.Command {
-	command := schema.Command{}
-	command.Id = GetRandomUniqueString(8, true)
-	LogInfoMessage(fmt.Sprintf("command Id: %s", command.Id))
+// AddCommand creates a command of a specified type in a schema structure and pupulates it with random attributes
+func (devfile *TestDevfile) AddCommand(commandType schema.CommandType) schema.Command {
 
+	var command *schema.Command
```
yangcao77, line 69: perhaps create type of `schema.Command`, and pass in `&command`
can use a switch case here. depends on your preference

tests/api/parser_v200_verify_test.go
```diff
@@ -80,27 +56,6 @@ func Test_CompositeCommand(t *testing.T) {
...
-	runTest(testContent, t)
-	runMultiThreadTest(testContent, t)
-}
-
-func Test_CompositeCommandEditParserCreate(t *testing.T) {
-	testContent := TestContent{}
-	testContent.CommandTypes = []schema.CommandType{schema.CompositeCommandType}
-	testContent.CreateWithParser = true
 	testContent.EditContent = true
 	testContent.FileName = utils.GetDevFileName()
```
yangcao77, line 60: minor suggestion: 
I saw all the tests created are setting those values. can create a func `newTestContent()`
```
newTestContent(commandType, editContent) {
	testContent := TestContent{}
	testContent.CommandTypes = commandType
	testContent.EditContent = editContent
	testContent.FileName = utils.GetDevFileName()

        return testContent
}

```

yangcao77:
Martim's test changes overall look good, just some minor suggestions.

tests/utils/command_test_utils.go
mmulholla: I kind of like it how it is but can looks at it more. OK to defer to later?

tests/api/parser_v200_verify_test.go
mmulholla: The kicker here is that utils.GetDevFileName() creates a filename based on the calling function name so it needs to be in the test function. I could update it to walk the stack until it finds the Test_* function but would rather not in this pull request.  Also I have though of similar in the past but just thought it would not save much.

mmulholla:
@elsony are we good to merge now? Can you approve?

elsony (approved):
/lgtm

PR #59, opened 2021-02-10
Restructure test directories and improve temp dir error handling

mmulholla:
1. Restructured the directories.
2. Restructure required a minor change to the tmp directory processing, so I added extra error checking to temp directory processing based on previous comments. Basically now if the temp directory can't be created the current directory will be used as temp. Also if the log file fails to be created the log function will print to the console instead.
3. Note CreateTempDir is external - changes coming require that


tests/v2/utils/test_utils.go
```diff
@@ -55,14 +60,23 @@ func init() {
...
+func CreateTempDir(subdir string) string {
 	tempDir := tmpDir + subdir + "/"
 	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
-		os.Mkdir(tempDir, 0755)
+		if err = os.Mkdir(tempDir, 0755); err != nil {
+			LogErrorMessage(fmt.Sprintf("Failed to create temp directory %s will use %s : %v", tempDir, tmpDir, err))
+			// if cannot create subdirectory just use the base tmp directory
+			tempDir = tmpDir
+		}
+	} else if err != nil {
```
yangcao77, line 76: why we need this extra else statement?
can pre-define the error

```
var err error
	if _, err = os.Stat(tempDir); os.IsNotExist(err) {
		err = os.Mkdir(tempDir, 0755)
	}
 if err != nil {
			LogErrorMessage(fmt.Sprintf("Failed to create temp directory %s will use %s : %v", tempDir, tmpDir, err))
			// if cannot create subdirectory just use the base tmp directory
			tempDir = tmpDi
```

mmulholla: with your example if os.IsNotExist(err) returned true your second if would also return true which would be a bug.  The else statement deals with errors which are not the  os.IsNotExist(err) for example a permission error.

yangcao77: The second won't return true if `os.Mkdir`  passes, err will be nil in this case

mmulholla: ah you removed my logic processing the os.Mkdir error, see where you are coming from now. Let me update it.

mmulholla: Uodated

yangcao77:
change looks good to me

maysunfaisal (approved):

PR #64, opened 2021-02-25
Automatic update schema

yangcao77:
### What does this PR do?
automate the process to adopt the latest devfile/api pkg, and the latest schema
detail can be found in: https://github.com/devfile/api/issues/358

Fixes https://github.com/devfile/api/issues/358

### DIFF
--- a/replaceSchemaFile.go
+++ b/replaceSchemaFile.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+func ReplaceSchemaFile() {
+	if len(os.Args) != 7 {
+		printErr(fmt.Errorf("ReplaceSchemaFile() expect 7 arguments"))
+		os.Exit(1)
+	}
+	originalSchema := os.Args[2]
+	schemaURL := os.Args[3]
+	packageVersion := os.Args[4]
+	jsonSchemaVersion := os.Args[5]
+	filePath := os.Args[6]
+
+	// replace all ` with ' to convert schema content from json file format to json format in golang
+	newSchema := strings.ReplaceAll(originalSchema, "`", "'")
+	fmt.Printf("Writing to file: %s\n", filePath)
+	fileContent := fmt.Sprintf("package %s\n\n// %s\nconst %s = `%s\n`\n", packageVersion, schemaURL, jsonSchemaVersion, newSchema)
+
+	if err := ioutil.WriteFile(filePath, []byte(fileContent), 0755); err != nil {
+		printErr(err)
+		os.Exit(1)
+	}
+}
+
+func printErr(err error) {
+	// prints error in red
+	colorRed := "\033[31m"
+	colorReset := "\033[0m"
+
+	fmt.Println(string(colorRed), err, string(colorReset))
+}
--- a/updateApi.sh
+++ b/updateApi.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+BLUE='\033[1;34m'
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+NC='\033[0m'
+BOLD='\033[1m'
+
+set -e
+
+CURRENT_DIR=$(pwd)
+API_PKG="github.com/devfile/api/v2"
+SCHEMA_URL_MASTER="https://raw.githubusercontent.com/devfile/api/master/schemas/latest/devfile.json"
+
+# 2.0.0 devfile
+SCHEMA_URL_200="https://raw.githubusercontent.com/devfile/api/2.0.x/schemas/latest/devfile.json"
+PACKAGE_VERSION_200="version200"
+JSON_SCHEMA_200="JsonSchema200"
+FILE_PATH_200="./pkg/devfile/parser/data/v2/2.0.0/devfileJsonSchema200.go"
+
+# 2.1.0 devfile
+PACKAGE_VERSION_210="version210"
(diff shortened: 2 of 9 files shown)


replaceSchemaFile.go
```diff
@@ -0,0 +1,26 @@
...
+	schemaURL := os.Args[3]
+
+	newSchema := strings.ReplaceAll(originalSchema, "`", "'")
+	fmt.Printf("Writing to file: %s\n", filePath)
+	fileContent := fmt.Sprintf("package %s\n\n// %s\nconst %s = `%s\n`\n", packageVersion, schemaURL, jsonSchemaVersion, newSchema)
+
+	if err := ioutil.WriteFile(filePath, []byte(fileContent), 0755); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
```
maysunfaisal, line 25: Is this only to meant to update current schema? How about 2.2.0? I think we also need to update the const in versions.go for 2.2.0 and may not be possible 🤔  and does that mean new version is always manual?

updateApi.sh
```diff
@@ -0,0 +1,31 @@
...
+BLUE='\033[1;34m'
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+NC='\033[0m'
+BOLD='\033[1m'
+
+set -e
+
+CURRENT_DIR=$(pwd)
+BASE_DIR=$(cd "$(dirname "$0")" && pwd)
```
maysunfaisal, line 12: dont think this is being used

updateApi.sh
```diff
@@ -0,0 +1,31 @@
...
+API_PKG="github.com/devfile/api/v2"
+SCHEMA_URL="https://raw.githubusercontent.com/devfile/api/master/schemas/latest/devfile.json"
+
+onError() {
+  cd "${CURRENT_DIR}"
+}
+trap 'onError' ERR
+
+
+echo "Updating devfile/api in go.mod"
```
maysunfaisal, line 22: can you use the color codes in your echo statements

updateApi.sh
```diff
@@ -0,0 +1,31 @@
...
+RED='\033[0;31m'
+NC='\033[0m'
+BOLD='\033[1m'
+
+set -e
+
+CURRENT_DIR=$(pwd)
+BASE_DIR=$(cd "$(dirname "$0")" && pwd)
+API_PKG="github.com/devfile/api/v2"
+SCHEMA_URL="https://raw.githubusercontent.com/devfile/api/master/schemas/latest/devfile.json"
```
maysunfaisal, line 14: I know its not frequent, but we may also want to hardcode and store 2.1.0 path since its never gonna change and there is a possibility that 2.1.0 schema maybe updated and in that case we can just pass input to this file to update the schema in our repo

updateApi.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "Updating devfile/api in go.mod"
+go get "${API_PKG}@master"
+
+echo "Get latest schema"
+schema=$(curl -L "${SCHEMA_URL}")
+
+#replace all ` with ' and write to schema file
+echo "Write to go file"
+go build *.go
+./main updateSchema "${schema}" "${SCHEMA_URL}"
```
maysunfaisal, line 31: I think we can also put this in our github action, so that every PR is always updated and we dont need to kick off the script. 

If there is a breaking change, our tests should catch it. WDYT?

updateApi.sh
maysunfaisal: If we go down that path its probably easier to maintain the logic in go that bash, where the bash is just calls the go exec with the params

replaceSchemaFile.go
yangcao77: For new versions, we need to set up once and update the const value in this file

updateApi.sh
yangcao77: 2.1.0 path is the master one. Do you mean 2.0.0 path?

maysunfaisal: yes sorry, 2.0.0 and so on when newer ones are out.

updateApi.sh
yangcao77: I was thinking our put it in github action whenever the schema has been updated. However, some schema changes would need library change accordingly, or otherwise would cause failure. We need to discuss more on that if we want to use github action

replaceSchemaFile.go
maysunfaisal: ok, then we can store paths, schema info and share logic with this https://github.com/devfile/library/pull/64#discussion_r585023117?

yangcao77: store all info in `./updateApi.sh`, and allow to pass an extra argument to update schema for selected devfile version
(now only has case 2.0.0, 2.1.0 will be default case)

updateApi.sh
yangcao77: removed

updateApi.sh
yangcao77: added color green for echo statements, and color red for errors

updateApi.sh
yangcao77: store all info in `./updateApi.sh`, and allow to pass an extra argument to update schema for selected devfile version
(now only has case 2.0.0, 2.1.0 will be default case)

updateApi.sh
maysunfaisal: yeah i see your point, in that case; when the github action opens a PR and the test fails because the new schema introduced breaking change, we can manually update that PR opened by the bot 🤔

maysunfaisal (approved):

PR #65, opened 2021-03-01
Add Delete Comp, Command, Proj, StarterProj

maysunfaisal:

### What does this PR do?
Adds ability to delete other devfile objects. As a result few items have changed:
- AddVolume() and DeleteVolume() are removed, AddComponents() and DeleteComponent() to be used
- Caller should also now call AddVolumeMount() and DeleteVolumeMount() to delete volume mounts when adding/removing volumes with AddComponents() and DeleteComponent() 
- Delete functions just removes the specified devfile object. References are not cleaned up ie; 
  - when a volume component is deleted, volume mounts are not removed
  - when a component is deleted, the command referencing the component are not dealt with
  - when a command is deleted, events referencing the command are not cleaned up
- Few interface func signature changes
  - `AddCommands(commands ...v1.Command) error` updated to `AddCommands(commands []v1.Command) error` to be consistent
  - `AddVolume(volume v1.Component, path string) error` removed
  - `	DeleteVolumeMount(name string) error` removed
  - `GetVolumeMountPath(name string) (string, error)` updated to `GetVolumeMountPath(mountName, componentName string) (string, error)` since a container can have multiple volume mounts and returning the first volume mount with the name is not logically correct
- Update Event funcs to handle pointers

Fixes https://github.com/devfile/api/issues/357
Fixes https://github.com/devfile/api/issues/363

yes, new tests


yangcao77:
Since delete functions are being added, I'm thinking if we should do validation after delete

For example, if a volume component, which is being referenced in a container component, got deleted, the new devfile becomes invalid. 

Same for the add & update functions

dharmit:
Commenting purely from `odo service delete` point of view. I was successfully able to delete the kubernetes component from the devfile. :+1:

maysunfaisal:
@yangcao77 When deleting a component we make sure the volume mounts are also deleted if its a volume. Similarly for command, we make sure the composite sub commands are also deleted if it matches the command to be deleted.

As for if it needs to be validated, I think we can make a call to Validate() func after writing to YAML. Since this is an exported func, tools like odo can also call this and this may also help ensure the data being written is validated. But it maybe overkill because we anyways validate when parsing the devfile which happens during every operation. WDYT

pkg/devfile/parser/data/interface.go
```diff
@@ -26,26 +26,30 @@ type DevfileData interface {
...
+	AddCommands(commands []v1.Command) error
 	UpdateCommand(command v1.Command)
+	DeleteCommand(id string) error
 
-	// volume related methods
-	AddVolume(volume v1.Component, path string) error
-	DeleteVolume(name string) error
-	GetVolumeMountPath(name string) (string, error)
+	// volume mount related methods
+	AddVolumeMount(componentName, name, path string) error
```
yangcao77, line 50:
```suggestion
	AddVolumeMount(componentName string, volumeMount v1.volumeMount) error
```

pkg/devfile/parser/data/v2/commands.go
```diff
@@ -57,3 +53,24 @@ func (d *DevfileV2) UpdateCommand(command v1.Command) {
 		}
 	}
 }
+
+// DeleteCommand removes the specified command
+func (d *DevfileV2) DeleteCommand(id string) error {
+
+	found := false
+	for i := len(d.Commands) - 1; i >= 0; i-- {
```
yangcao77, line 61: can use `for index := range d.Commands` to find the index
and `append(d.Commands[:index], slice[index+1:]...)`

pkg/devfile/parser/data/v2/commands_test.go
```diff
@@ -300,3 +301,102 @@ func TestDevfile200_UpdateCommands(t *testing.T) {
...
+
+	tests := []struct {
+		name            string
+		commandToDelete string
+		commands        []v1.Command
+		wantCommands    []v1.Command
+		wantErr         bool
+	}{
+		{
+			name:            "Commands that belong to Composite Command",
```
yangcao77, line 315: rename the test case to avoid confusion, as composite command is not a special case now

pkg/devfile/parser/data/v2/commands_test.go
```diff
@@ -300,3 +301,102 @@ func TestDevfile200_UpdateCommands(t *testing.T) {
...
+						Exec: &v1.ExecCommand{},
+					},
+				},
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			d := &DevfileV2{
```
yangcao77, line 381: can we have the devObj defined outside? so we do not need to define commands in each of the test case.
and the test args can be 
```
	tests := []struct {
		name            string
		commandToDelete string
		wantCommands    []v1.Command
		wantErr         bool
	}
```

pkg/devfile/parser/data/v2/commands_test.go
```diff
@@ -300,3 +301,102 @@ func TestDevfile200_UpdateCommands(t *testing.T) {
...
+			commandToDelete: "command34",
+			commands: []v1.Command{
+				{
+					Id: "command1",
+					CommandUnion: v1.CommandUnion{
+						Exec: &v1.ExecCommand{},
+					},
+				},
+			},
+			wantCommands: []v1.Command{
```
yangcao77, line 368: For the error case, looks like that `wantCommands` is not necessary

pkg/devfile/parser/data/v2/components_test.go
```diff
@@ -403,3 +404,114 @@ func TestGetDevfileVolumeComponents(t *testing.T) {
...
+			componentToDelete: "comp12",
+			components: []v1.Component{
+				{
+					Name: "comp2",
+					ComponentUnion: v1.ComponentUnion{
+						Volume: &v1.VolumeComponent{},
+					},
+				},
+			},
+			wantComponents: []v1.Component{
```
yangcao77, line 483: For the error case, looks like that `wantComponents` is not necessary

pkg/devfile/parser/data/v2/projects_test.go
```diff
@@ -188,6 +189,83 @@ func TestDevfile200_UpdateProject(t *testing.T) {
...
+				{
+					Name:      "nodejs",
+					ClonePath: "/project",
+				},
+				{
+					Name:      "java",
+					ClonePath: "/project2",
+				},
+			},
+			wantProjects: []v1.Project{
```
yangcao77, line 235: for the error case, `wantProjects` do not need to be defined

pkg/devfile/parser/data/v2/components_test.go
```diff
@@ -403,3 +404,114 @@ func TestGetDevfileVolumeComponents(t *testing.T) {
...
+
+	tests := []struct {
+		name              string
+		componentToDelete string
+		components        []v1.Component
+		wantComponents    []v1.Component
+		wantErr           bool
+	}{
+		{
+			name:              "Volume Component with mounts",
```
yangcao77, line 418: since we do pure deletion without considering the reference, we should rename the test case to avoid confusion

pkg/devfile/parser/data/v2/projects_test.go
```diff
@@ -370,3 +448,80 @@ func TestDevfile200_UpdateStarterProject(t *testing.T) {
...
+				{
+					Name:   "nodejs",
+					SubDir: "/project",
+				},
+				{
+					Name:   "java",
+					SubDir: "/project2",
+				},
+			},
+			wantStarterProjects: []v1.StarterProject{
```
yangcao77, line 495: for the error case, wantStarterProjects do not need to be defined

pkg/devfile/parser/data/v2/volumes.go
```diff
@@ -8,99 +8,90 @@ import (
 	"github.com/devfile/library/pkg/devfile/parser/data/v2/common"
 )
 
-// AddVolume adds the volume to the devFile and mounts it to all the container components
-func (d *DevfileV2) AddVolume(volumeComponent v1.Component, path string) error {
-	volumeExists := false
+// AddVolumeMount adds the volume mount to the specified container component
+func (d *DevfileV2) AddVolumeMount(componentName, name, path string) error {
```
yangcao77, line 12: since other APIs take go struct if it's ever defined in devfile api, suggest to use `v1.VolumeMount` to be consistent with other APIs

pkg/devfile/parser/data/v2/volumes_test.go
```diff
@@ -175,248 +155,157 @@ func TestDevfile200_AddVolume(t *testing.T) {
...
-			}
-
-			if err != nil && tt.wantErr {
-				return
-			}
-
-			if !reflect.DeepEqual(d.Components, tt.wantComponents) {
-				t.Errorf("wanted: %v, got: %v, difference at %v", tt.wantComponents, d.Components, pretty.Compare(tt.wantComponents, d.Components))
+			err := d.AddVolumeMount(tt.args.componentName, tt.args.name, tt.args.path)
+			if tt.wantErr && err == nil {
```
yangcao77, line 159: why split this logic `(err != nil) != tt.wantErr` into two if-else blocks? 
the existing one is clear enough to me

Same for other same changes.

pkg/devfile/parser/data/v2/volumes_test.go
```diff
@@ -175,248 +155,157 @@ func TestDevfile200_AddVolume(t *testing.T) {
...
+				t.Errorf("Expected error from test but got nil")
+			} else if !tt.wantErr && err != nil {
+				t.Errorf("Got unexpected error: %s", err)
+			} else if err == nil {
+				assert.Equal(t, tt.wantComponents, d.Components, "The two values should be the same.")
```
yangcao77, line 164: I prefer the existing assertion, with difference printed out. This will make debugging more difficult

yangcao77 (changes requested):

pkg/devfile/parser/data/v2/components.go
```diff
@@ -88,3 +84,24 @@ func (d *DevfileV2) UpdateComponent(component v1.Component) {
 		d.Components[index] = component
 	}
 }
+
+// DeleteComponent removes the specified component
+func (d *DevfileV2) DeleteComponent(name string) error {
+
+	found := false
+	for i := len(d.Components) - 1; i >= 0; i-- {
```
yangcao77, line 92: same as https://github.com/devfile/library/pull/65/files#r587556660

pkg/devfile/parser/data/v2/projects.go
```diff
@@ -55,6 +55,27 @@ func (d *DevfileV2) UpdateProject(project v1.Project) {
 	}
 }
 
+// DeleteProject removes the specified project
+func (d *DevfileV2) DeleteProject(name string) error {
+
+	found := false
+	for i := len(d.Projects) - 1; i >= 0; i-- {
```
yangcao77, line 62: https://github.com/devfile/library/pull/65/files#r587556660

yangcao77:
Also, can we add description in the PR's main description, stating that all deletion APIs are just doing basic deletion ,the returned devfileObj is not guaranteed to be valid?

I saw volume mount APIs got changed a lot, probably also want to state those changes in the main description, so odo folks can easily get what have been changed; and what is expected now from those functions
for example, `GetVolumeMountPath` now only finds out if the volume mount is defined, and return the path; it's not guaranteed that the volume component is defined.

pkg/devfile/parser/data/interface.go
```diff
@@ -26,26 +26,30 @@ type DevfileData interface {
...
 	// starter projects related commands
 	GetStarterProjects(common.DevfileOptions) ([]v1.StarterProject, error)
 	AddStarterProjects(projects []v1.StarterProject) error
 	UpdateStarterProject(project v1.StarterProject)
+	DeleteStarterProject(name string) error
 
 	// command related methods
 	GetCommands(common.DevfileOptions) ([]v1.Command, error)
-	AddCommands(commands ...v1.Command) error
+	AddCommands(commands []v1.Command) error
```
kadel, line 45: This is changing the signature of an existing function. It will break every project that already uses this function.

pkg/devfile/parser/data/interface.go
```diff
@@ -26,26 +26,30 @@ type DevfileData interface {
...
 	UpdateCommand(command v1.Command)
+	DeleteCommand(id string) error
 
-	// volume related methods
-	AddVolume(volume v1.Component, path string) error
-	DeleteVolume(name string) error
```
kadel, line 47: This is removing existing functions that other tools might be already using. This is a breaking change.

kadel (changes requested):
This PR is introducing a bunch of breaking changes.
What is the versioning policy for this library?

How other projects using this library can know when it is safe to update the library and when it is not because there were breaking changes in the interface? 

UPDATE:
That might sound maybe a little bit harsh. :-( I don't want to say that API can't change in a breaking way. But in order for people to use any library there need to be clear version rules and ideally a changelog so consumers can know what changed when.

pkg/devfile/parser/data/interface.go
maysunfaisal: @kadel Thanks for the review and concern. Yes, you're right, there should be a release every sprint, specifically when there is a breaking change with the changelog of the PR/commit.

Let me think about it a bit more. Since devfile/library has never had a release, I can do a v1.0.0-alpha release after this PR.

What do you think?

pkg/devfile/parser/data/interface.go
maysunfaisal: you're right

pkg/devfile/parser/data/interface.go
maysunfaisal: I will make it an array `AddVolumeMount(componentName string, volumeMounts []v1.volumeMount) error` since all existing add are arrays

pkg/devfile/parser/data/v2/volumes_test.go
maysunfaisal: I prefer the assert module, other devfile repos are using it and the output is pretty clean and detailed with less t.Errorf print statements:

```
--- FAIL: TestDevfile200_AddVolumeMount/add_the_volume_mount_when_other_mounts_are_present (0.00s)
        /Users/maysun/dev/devfile/parser/pkg/devfile/parser/data/v2/volumes_test.go:172: 
            	Error Trace:	volumes_test.go:172
            	Error:      	Not equal: 
            	            	expected: []v1alpha2.Component{v1alpha2.Component{Name:"container0", Attributes:attributes.Attributes(nil), ComponentUnion:v1alpha2.ComponentUnion{ComponentType:"", Container:(*v1alpha2.ContainerComponent)(0xc0001a01e0), Kubernetes:(*v1alpha2.KubernetesComponent)(nil), Openshift:(*v1alpha2.OpenshiftComponent)(nil), Volume:(*v1alpha2.VolumeComponent)(nil), Plugin:(*v1alpha2.PluginComponent)(nil), Custom:(*v1alpha2.CustomComponent)(nil)}}, v1alpha2.Component{Name:"container1", Attributes:attributes.Attributes(nil), ComponentUnion:v1alpha2.ComponentUnion{ComponentType:"", Container:(*v1alpha2.ContainerComponent)(0xc0001a02d0), Kubernetes:(*v1alpha2.KubernetesComponent)(nil), Openshift:(*v1alpha2.OpenshiftComponent)(nil), Volume:(*v1alpha2.VolumeComponent)(nil), Plugin:(*v1alpha2.PluginComponent)(nil), Custom:(*v1alpha2.CustomComponent)(nil)}}}
            	            	actual  : []v1alpha2.Component{v1alpha2.Component{Name:"container0", Attributes:attributes.Attributes(nil), ComponentUnion:v1alpha2.ComponentUnion{ComponentType:"", Container:(*v1alpha2.ContainerComponent)(0xc0001a0000), Kubernetes:(*v1alpha2.KubernetesComponent)(nil), Openshift:(*v1alpha2.OpenshiftComponent)(nil), Volume:(*v1alpha2.VolumeComponent)(nil), Plugin:(*v1alpha2.PluginComponent)(nil), Custom:(*v1alpha2.CustomComponent)(nil)}}, v1alpha2.Component{Name:"container1", Attributes:attributes.Attributes(nil), ComponentUnion:v1alpha2.ComponentUnion{ComponentType:"", Container:(*v1alpha2.ContainerComponent)(0xc0001a00f0), Kubernetes:(*v1alpha2.KubernetesComponent)(nil), Openshift:(*v1alpha2.OpenshiftComponent)(nil), Volume:(*v1alpha2.VolumeComponent)(nil), Plugin:(*v1alpha2.PluginComponent)(nil), Custom:(*v1alpha2.CustomComponent)(nil)}}}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -15,3 +15,3 @@
            	            	        Name: (string) (len=7) "volume1",
            	            	-       Path: (string) (len=3) "/da"
            	            	+       Path: (string) (len=5) "/data"
            	            	       },
            	Test:       	TestDevfile200_AddVolumeMount/add_the_volume_mount_when_other_mounts_are_present
            	Messages:   	The two values should be the same.
 ```

File: pkg/devfile/parser/data/interface.go (hunk 15 above)
Comment (kadel) [reply pkg/devfile/parser/data/interface.go:45]: That would be great. At least we would know when and what problems we can expect when upgrading to a newer version.

File: pkg/devfile/parser/data/v2/commands.go [hunk 17]
```diff
@@ -57,3 +53,25 @@ func (d *DevfileV2) UpdateCommand(command v1.Command) {
...
+
+// DeleteCommand removes the specified command
+func (d *DevfileV2) DeleteCommand(id string) error {
+
+	found := false
+	for i := range d.Commands {
+		if d.Commands[i].Id == id {
+			found = true
+			d.Commands = append(d.Commands[:i], d.Commands[i+1:]...)
+			break
```
yangcao77, line 65: actually can just `return nil` at this point. and the `found` is not needed
same for the other delete functions

pkg/devfile/parser/data/v2/commands_test.go
```diff
@@ -300,3 +301,84 @@ func TestDevfile200_UpdateCommands(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := d.DeleteCommand(tt.commandToDelete)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("DeleteCommand() error = %v, wantErr %v", err, tt.wantErr)
+			}
+
+			if err == nil && !tt.wantErr {
+				assert.Equal(t, tt.wantCommands, d.Commands, "The two values should be the same.")
```
yangcao77, line 379: this if statement, already checked by the previous one (previous if block should return)
can directly check tt.wantCommands &  d.Commands
same for the other tests

pkg/devfile/parser/data/v2/events_test.go
```diff
@@ -50,17 +60,17 @@ func TestDevfile200_AddEvents(t *testing.T) {
...
 				},
 			}
 
-			got := d.AddEvents(tt.newEvents)
+			err := d.AddEvents(tt.newEvents)
 
-			if !tt.wantErr && got != nil {
-				t.Errorf("TestDevfile200_AddEvents() unexpected error - %+v", got)
-			} else if tt.wantErr && got == nil {
+			if !tt.wantErr && err != nil {
```
yangcao77, line 71: we can combine this two checks into one `if (err != nil) != tt.wantErr `

pkg/devfile/parser/data/v2/volumes.go
```diff
@@ -8,99 +8,90 @@ import (
...
 	found := false
-	for i := len(d.Components) - 1; i >= 0; i-- {
-		if d.Components[i].Container != nil {
-			var tmp []v1.VolumeMount
-			for _, volumeMount := range d.Components[i].Container.VolumeMounts {
-				if volumeMount.Name != name {
-					tmp = append(tmp, volumeMount)
+	for i := range d.Components {
+		if d.Components[i].Container != nil && d.Components[i].Name != name {
+			for j := len(d.Components[i].Container.VolumeMounts) - 1; j >= 0; j-- {
```
yangcao77, line 50: we should have a comment here explains why we are doing a decremental for loop

pkg/devfile/parser/data/v2/volumes.go
```diff
@@ -8,99 +8,90 @@ import (
...
 
 	for _, component := range d.Components {
-		if component.Container != nil {
+		if component.Container != nil && component.Name == componentName {
+			componentFound = true
 			for _, volumeMount := range component.Container.VolumeMounts {
-				if volumeMount.Name == name {
+				if volumeMount.Name == mountName {
 					mountFound = true
 					path = volumeMount.Path
```
yangcao77, line 81: directly return  if component & mount both found

yangcao77 (changes requested):

pkg/devfile/parser/data/v2/commands_test.go
maysunfaisal: no we cant, because there are tests where we want an err and that should be skipped. The prev if block is to check if there is an err mis match. I've updated it to a better if else block

yangcao77: it won't be skipped, because it's returning from that single test `func(t *testing.T){}`, it will continue the for loop for the rest of tests

pkg/devfile/parser/data/interface.go
maysunfaisal: @kadel do you have any other concerns before I merge this PR? I am looking to do a pre-release once this PR is in and hopefully generate a changelog for the release so its easier to track changes.

PR #68, opened 2021-03-05
update to use EPL-2.0 License

yangcao77:

### What does this PR do?
This PR updates the project to use EPL-2.0 License to prepare for release


elsony:
Given that it is unclear that we'll eventually use Apache or EPL license at this point. We should hold off from committing this change until we have the final decision.

PR #74, opened 2021-03-19
Add library package for resolving devfile parents and plugins

amisevsk:
### What does this PR do?
This PR is a port of the [`pkg/library/flatten`](https://pkg.go.dev/github.com/devfile/devworkspace-operator/pkg/library/flatten) library in the DevWorkspace operator. I initially wrote it to quickly add support for plugins in the DevWorkspace Operator, but it should be useful to anyone working with devfiles. 

This PR starts with a relatively straightforward port of the current work in DWO. For additional context in developing this, see PRs https://github.com/devfile/devworkspace-operator/pull/240 and https://github.com/devfile/devworkspace-operator/pull/269.

The basic API defined here is
```go
func ResolveDevWorkspace(workspace *devfile.DevWorkspaceTemplateSpec, tooling ResolverTools) (*devfile.DevWorkspaceTemplateSpec, error) 
```
which takes a `DevWorkspaceTemplateSpec` and returns the equivalent `DevWorkspaceTemplateSpec` with all of the plugins and parents resolved and merged/overridden. The `ResolverTools` struct contains all the interfaces needed to actually grab plugins from outside the current binary:
```go
type ResolverTools struct {
	// DefaultNamespace is the default namespace to use for resolving Kubernetes ImportReferences that do not include one
	DefaultNamespace string
	// DefaultRegistryURL is the default registry URL to use when a component specifies an id but not registryURL
	DefaultRegistryURL string
	// Context is the context used for making Kubernetes or HTTP requests
	Context context.Context
	// K8sClient is the Kubernetes client instance used for interacting with a cluster
	K8sClient client.Client
	// HttpClient is the HTTP client used for making network requests when resolving plugins or parents.
	HttpClient network.HTTPGetter
}
```
If any field is `nil` or empty, an error is returned only if it is used. This means using this library to support *only* Kubernetes-type references is as simple as providing no HttpClient and checking the error returned (this could probably be improved with error types).

One issue we ran into in DWO is that it can be hard to determine where e.g. a component in the flattened devfile came from. To help with this, the merge process adds attributes `library.devfile.io/imported-by=source` to Components, Commands, Projects, and StarterProjects that are not in the original devfile. For plugins `<source>` is the name of the plugin component that imported them, for parents, source is `parent`.

The code is fairly well-tested (coverage >90%), and there's not actually _that_ many non-test lines of code here (~550 lines of actual code, the rest is tests).

I've tried to keep the commit history in this PR readable commit-by-commit (with helpful commit messages), and tried to keep test cases separate from functional changes.

I don't know if there is an issue for this

We use this code regularly in the DevWorkspace Operator without issue. Running tests on the changes directly can be done by executing
```
go test ./pkg/flatten/...
```


amisevsk:
Some additional thoughts on this PR:
- The approach here may be complicated by https://github.com/devfile/api/pull/333 since we'll likely not have a common struct (`DevWorkspaceTemplateSpec`) to represent both the spec of a devfile and DevWorkspace
- The devworkspace operator currently uses the EPL 2.0 license, we need to figure out how to reconcile this with Apache (I believe we want to switch DWO to Apache as well). Files existing in DWO currently have an EPL license header that will probably need to be removed.

yangcao77:
Since plugins is going to be removed from devfile by https://github.com/devfile/api/pull/333, we decided not to add any more plugins support into devfile/library. 

In addition, the parser in library now contains it's own logic to get flattened devfile, and support parent reference using uri & id. It supports more than what DWO currently has, i.e. [relative uri](https://github.com/devfile/library/pull/57)(local path & URL), and [multiple registry URLs](https://github.com/devfile/library/blob/64d0bf9cc438090eaab04f6a77c2c42d22d97cca/pkg/devfile/parser/parse.go#L69-L71) can be passed in to parser. 

Kubernetes CRD is the one we want to add support to. We might take some pieces from this PR, which resolves Kubernetes CRD reference, and combine with the existing logic

amisevsk:
> Since plugins is going to be removed from devfile by devfile/api#333, we decided not to add any more plugins support into devfile/library

Well, plugins are being removed from the devfile, but they are staying in the DevWorkspace spec for now.

> In addition, the parser in library now contains it's own logic to get flattened devfile, and support parent reference using uri & id. It supports more than what DWO currently has, i.e. relative uri(local path & URL), and multiple registry URLs can be passed in to parser.

I think our set of needs is different :) -- there's no relative URIs in a controller.

pkg/flatten/flatten.go
```diff
@@ -0,0 +1,268 @@
...
+	// K8sClient is the Kubernetes client instance used for interacting with a cluster
+	K8sClient client.Client
+	// HttpClient is the HTTP client used for making network requests when resolving plugins or parents.
+	HttpClient network.HTTPGetter
+}
+
+// ResolveDevWorkspace takes a DevWorkspaceTemplateSpec and returns a "resolved" version of it -- i.e. one where all plugins and parents
+// are inlined as components.
+// TODO:
+// - Implement flattening for DevWorkspace parents
```
sleshchenko, line 46: I see `resolveParentComponent` func, isn't it already done?

pkg/flatten/testdata/annotate/annotate-devfile-with-importing-source.yaml
```diff
@@ -0,0 +1,68 @@
...
+        container:
+          image: plugin-img
+    commands:
+      - id: parent-command
+        attributes:
+          library.devfile.io/imported-by: parent
+      - id: plugin-command
+        attributes:
+          library.devfile.io/imported-by: test-plugin
+
```
sleshchenko, line 68:
```suggestion
```

pkg/flatten/testdata/general/fail-nicely-when-no-http-client-provided_id.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+input:
+  workspace:
+    components:
+      - name: test-plugin
+        plugin:
+          id: test/plugin
+
+output:
+  errRegexp: "cannot resolve resources by id: no HTTP client provided"
```
sleshchenko, line 11:
```suggestion
  errRegexp: "cannot resolve resources by id: no HTTP client provided"

```
the value of my review increases after I commented all end-file empty lines )
Here and a few places below.

pkg/flatten/flatten.go
```diff
@@ -0,0 +1,268 @@
...
+func ResolveDevWorkspace(workspace *devfile.DevWorkspaceTemplateSpec, tooling ResolverTools) (*devfile.DevWorkspaceTemplateSpec, error) {
```
sleshchenko, line 47: I assume it may simplify testing for clients if such functionality would be provided with interface. Like DevWorkspaceResolver.
Then here we would not need to pass resolverTooling which is good thing as well I think, instead we'll have something like:
```
r := resolver.NewWithTooling(tooling)
r.resolveDevWorkspace(tooling)
```

But it's not a strong suggestion, just assumption.

sleshchenko (commented):
The code LGTM.

A bit later will do another round of review to take a look test cases as well.

sleshchenko:
> A bit later will do another round of review to take a look test cases as well.

I'm taking my words back ) I don't plan to do more detailed review since it does not seem this PR will be merged due plugins. If it's changed, please request my review one more time.

go.mod
```diff
@@ -7,18 +7,23 @@ require (
...
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/gobwas/glob v0.2.3
+	github.com/golang/groupcache v0.0.0-20191227052852-215e87163ea7 // indirect
 	github.com/google/go-cmp v0.4.0
 	github.com/gregjones/httpcache v0.0.0-20180305231024-9cad4c3443a7
 	github.com/kylelemons/godebug v0.0.0-20170820004349-d65d576e9348
 	github.com/mattn/go-colorable v0.1.2 // indirect
 	github.com/mattn/go-isatty v0.0.12 // indirect
-	github.com/openshift/api v0.0.0-20200930075302-db52bc4ef99f
+	github.com/openshift/api v3.9.0+incompatible
```
maysunfaisal, line 16: I think we should stick with openshift/api version 4.6 or higher. I had to explicitly update to a newer version because openshift/console and openshift/odo uses newer versions and there is a breaking dependency change when going from version 3.9 to 4.0 and above

yangcao77:
> Well, plugins are being removed from the devfile, but they are staying in the DevWorkspace spec for now.
 
Fair enough. So library is not going to support plugin from devfile spec perspective, but if the devWorkspaceTemplate specified under parent's kubernetes CRD, we will parse and put it in the flattened devfile. If DWO needs more plugin support, you will need to handle it in DWO specifically. Does that sound good to you?


> I think our set of needs is different :) -- there's no relative URIs in a controller.
The existing library also accepts absolute URI address as DWO expected, the relative URIs support is just an extra.

pkg/flatten/flatten.go
amisevsk: Forgot to remove the TODO after implementing :)

go.mod
amisevsk: Yeah, I'm not sure why this got pinned to `3.9.0` -- I assume it's Go's automatic "upgrading" (3.9 is bigger than 0.0?)

I'll fix this.

amisevsk:
Closing this PR as this is not the approach devfile/library will take.

PR #76, opened 2021-03-23
restricts registryURL format

yangcao77:

### What does this PR do?
This PR restricts the `registryURL` format, protocol is expected to be included as part of the `registryURL`

Since the filed is `registryURL`, it is expected to follow the url format. library does not need to help resolve the format, and should not take registry host as the url

Fixes https://github.com/devfile/api/issues/386

Update unit tests

### DIFF
--- a/pkg/devfile/parser/parse_test.go
+++ b/pkg/devfile/parser/parse_test.go
@@ -2978,7 +2978,7 @@ func Test_parseFromRegistry(t *testing.T) {
 		wantErr       bool
 	}{
 		{
-			name:          "should be able to parse from provided registryUrl without prefix",
+			name:          "should fail if provided registryUrl does not have protocol prefix",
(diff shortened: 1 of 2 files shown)


pkg/devfile/parser/parse_test.go
```diff
@@ -2983,6 +2983,7 @@ func Test_parseFromRegistry(t *testing.T) {
 			wantDevFile:   parentDevfile,
 			registryUrl:   registry,
 			registryId:    registryId,
+			wantErr:       true,
```
maysunfaisal, line 2986: test title should be `should NOT be able to parse from provided registryUrl without prefix`

yangcao77: updated

maysunfaisal (approved):

PR #77, opened 2021-03-30
remove file name addition to URL

yangcao77:

### What does this PR do?
Remove file name addition to URL, URL is expected to be a file path

Fixes https://github.com/devfile/api/issues/398

Updated unit test

### DIFF
--- a/pkg/devfile/parser/context/context.go
+++ b/pkg/devfile/parser/context/context.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"path"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


maysunfaisal (approved):
lgtm

curious, do we have tests with uri not ending in devfile or with devfile.yaml that we can update?

yangcao77:
This test is without devfile.yaml: https://github.com/devfile/library/blob/7ad86fe212a2973fd0fafc0befa212dede337c06/pkg/devfile/parser/parse_test.go#L2202
So at the end of the test, the expected error message is updated to be without appending devfile.yaml

PR #80, opened 2021-04-16
397 add annotation to imported elements

yangcao77:
### What does this PR do?
This PR adds an attribute to all elements from the imported source before override & merge. This attribute information can be used by validator to track from which devfile the error came from. 

https://github.com/devfile/api/issues/397

Update unit tests. Add more unit tests to test the import attribute fields


maysunfaisal (commented):
could you also update the line here where we dont check for err when parsing from CRD:

Screen Shot 2021-04-20 at 1 19 31 PM

pkg/devfile/parser/sourceAttribute.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// AddSourceAttributesForParentOverride adds an attribute 'library.devfile.io/imported-from=<source reference>'
+//  to all elements of parent override that support attributes.
+func AddSourceAttributesForParentOverride(sourceImportReference v1.ImportReference, parentoverride *v1.ParentOverrides) {
+	for idx, component := range parentoverride.Components {
+		if component.Attributes == nil {
+			parentoverride.Components[idx].Attributes = attributes.Attributes{}
+		}
+		parentoverride.Components[idx].Attributes.PutString(ImportSourceAttribute, fmt.Sprintf("parentOverrides from: %s", resolveImportReference(sourceImportReference)))
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/devfile/parser/sourceAttribute.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// AddSourceAttributesForPluginOverride adds an attribute 'library.devfile.io/imported-from=<source reference>'
+//  to all elements of plugin override that support attributes.
+func AddSourceAttributesForPluginOverride(sourceImportReference v1.ImportReference, pluginName string, pluginoverride *v1.PluginOverrides) {
+	for idx, component := range pluginoverride.Components {
+		if component.Attributes == nil {
+			pluginoverride.Components[idx].Attributes = attributes.Attributes{}
+		}
+		pluginoverride.Components[idx].Attributes.PutString(ImportSourceAttribute, fmt.Sprintf("pluginOverrides from: %s, plugin : %s", resolveImportReference(sourceImportReference), pluginName))
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/devfile/parser/sourceAttribute.go
```diff
@@ -0,0 +1,86 @@
...
+			template.StarterProjects[idx].Attributes = attributes.Attributes{}
+		}
+		template.StarterProjects[idx].Attributes.PutString(ImportSourceAttribute, resolveImportReference(sourceImportReference))
+	}
+}
+
+// AddSourceAttributesForParentOverride adds an attribute 'library.devfile.io/imported-from=<source reference>'
+//  to all elements of parent override that support attributes.
+func AddSourceAttributesForParentOverride(sourceImportReference v1.ImportReference, parentoverride *v1.ParentOverrides) {
+	for idx, component := range parentoverride.Components {
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/devfile/parser/sourceAttribute.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+	v1 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
+	"github.com/devfile/api/v2/pkg/attributes"
+)
+
+const ImportSourceAttribute = "library.devfile.io/imported-from"
+
+// AddSourceAttributesForTemplateSpecContent adds an attribute 'library.devfile.io/imported-from=<source reference>'
+//  to all elements of template spec content that support attributes.
+func AddSourceAttributesForTemplateSpecContent(sourceImportReference v1.ImportReference, template *v1.DevWorkspaceTemplateSpecContent) {
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/devfile/parser/parse_test.go
```diff
@@ -26,6 +27,74 @@ import (
 const schemaV200 = "2.0.0"
 
 func Test_parseParentAndPluginFromURI(t *testing.T) {
+	const uri1 = "127.0.0.1:8080"
+	const uri2 = "127.0.0.1:9090"
+	const pluginName = "plugincomp"
+	importFromUri1 := attributes.Attributes{}
+	importFromUri1.PutString(ImportSourceAttribute, fmt.Sprintf("uri: http://%s", uri1))
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/devfile/parser/parse.go
```diff
@@ -250,8 +255,13 @@ func parseParentAndPlugin(d DevfileObj, resolveCtx *resolutionContextTree, tool
 				return fmt.Errorf("plugin %s does not define any resources", component.Name)
 			}
 			pluginWorkspaceContent := pluginDevfileObj.Data.GetDevfileWorkspaceSpecContent()
+			// add attribute to plugin elements
+			AddSourceAttributesForTemplateSpecContent(plugin.ImportReference, pluginWorkspaceContent)
 			flattenedPlugin := pluginWorkspaceContent
 			if !reflect.DeepEqual(plugin.PluginOverrides, v1.PluginOverrides{}) {
+				// add attribute to parentOverrides elements
```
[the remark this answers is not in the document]

pkg/devfile/parser/sourceAttribute.go
```diff
@@ -37,50 +39,74 @@ func AddSourceAttributesForTemplateSpecContent(sourceImportReference v1.ImportRe
...
-		pluginoverride.Commands[idx].Attributes.PutString(ImportSourceAttribute, fmt.Sprintf("pluginOverrides from: %s, plugin : %s", resolveImportReference(sourceImportReference), pluginName))
+		pluginOverrides.Commands[idx].Attributes.PutString(PluginOverrideAttribute, resolveImportReference(sourceImportReference))
 	}
 
 }
+
+// AddSourceAttributesForOverrideAndMerge adds an attribute record the import reference to all elements of template that support attributes.
+func AddSourceAttributesForOverrideAndMerge(sourceImportReference v1.ImportReference, template interface{}) error {
+	if template == nil {
+		fmt.Errorf("cannot add source attributes to nil")
```
maysunfaisal, line 93: return

now i want us to test this use case as well 🙃  do you think you can fit it in your current test cases, if you dont want a new test for it

pkg/devfile/parser/sourceAttribute.go
```diff
@@ -37,50 +39,74 @@ func AddSourceAttributesForTemplateSpecContent(sourceImportReference v1.ImportRe
...
+			parentOverrides.StarterProjects[idx].Attributes = attributes.Attributes{}
 		}
-		parentoverride.StarterProjects[idx].Attributes.PutString(ImportSourceAttribute, fmt.Sprintf("parentOverrides from: %s", resolveImportReference(sourceImportReference)))
+		parentOverrides.StarterProjects[idx].Attributes.PutString(ParentOverrideAttribute, resolveImportReference(sourceImportReference))
 	}
 
 }
 
-// AddSourceAttributesForPluginOverride adds an attribute 'library.devfile.io/imported-from=<source reference>'
+// addSourceAttributesForPluginOverride adds an attribute 'library.devfile.io/imported-from=<source reference>'
```
maysunfaisal, line 72: need to update func desc for the updated attribute key, same for the parent override func

pkg/devfile/parser/sourceAttribute.go
```diff
@@ -37,50 +39,74 @@ func AddSourceAttributesForTemplateSpecContent(sourceImportReference v1.ImportRe
...
+			pluginOverrides.Commands[idx].Attributes = attributes.Attributes{}
 		}
-		pluginoverride.Commands[idx].Attributes.PutString(ImportSourceAttribute, fmt.Sprintf("pluginOverrides from: %s, plugin : %s", resolveImportReference(sourceImportReference), pluginName))
+		pluginOverrides.Commands[idx].Attributes.PutString(PluginOverrideAttribute, resolveImportReference(sourceImportReference))
 	}
 
 }
+
+// AddSourceAttributesForOverrideAndMerge adds an attribute record the import reference to all elements of template that support attributes.
+func AddSourceAttributesForOverrideAndMerge(sourceImportReference v1.ImportReference, template interface{}) error {
```
maysunfaisal, line 91: thinking about it, do we need this to be exported or can this remain internal as well? Since no one else will probably call it

pkg/devfile/parser/sourceAttribute.go
```diff
@@ -7,10 +7,12 @@ import (
 )
 
 const ImportSourceAttribute = "library.devfile.io/imported-from"
+const PluginOverrideAttribute = "library.devfile.io/plugin-override-from"
+const ParentOverrideAttribute = "library.devfile.io/parent-override-from"
```
maysunfaisal, line 11: should we keep this as unexported?

pkg/devfile/parser/parse.go
```diff
@@ -212,11 +212,11 @@ func parseParentAndPlugin(d DevfileObj, resolveCtx *resolutionContextTree, tool
 
 			parentWorkspaceContent := parentDevfileObj.Data.GetDevfileWorkspaceSpecContent()
 			// add attribute to parent elements
-			AddSourceAttributesForTemplateSpecContent(parent.ImportReference, parentWorkspaceContent)
+			AddSourceAttributesForOverrideAndMerge(parent.ImportReference, parentWorkspaceContent)
```
maysunfaisal, line 215: now we need to check for err

maysunfaisal:
@yangcao77 Can you also check at this screenshot? https://github.com/devfile/library/pull/80#pullrequestreview-640130084

yangcao77:
@maysunfaisal https://github.com/devfile/library/pull/80#pullrequestreview-640130084 has been fixed. the error is checked and returned outside of the switch case

pkg/devfile/parser/sourceAttribute.go
yangcao77: it is not going to fit in current test, as the template is never going to be `nil` when we call it from `parseParentAndPlugin`.

pkg/devfile/parser/sourceAttribute.go
yangcao77: yes, we can make it unexported. but I'm thinking if we should defined those const in devfile/api?? as we need to check those keys before return an error from validator.  WDYT?

maysunfaisal: I was meaning to ask, how would we check for these in the validation since DWO has their own library; so i dont think we can expect to that. This is will make devfile/api dependant on devfile/library...

If we're sure about the attribute keys and they're the same between devfile/library and DWO library, I think we can declare these in devfile/api

pkg/devfile/parser/sourceAttribute.go
yangcao77: updated

pkg/devfile/parser/sourceAttribute.go
yangcao77: updated

pkg/devfile/parser/sourceAttribute.go
yangcao77: updated

pkg/devfile/parser/sourceAttribute.go
yangcao77: yep. That's what I'm mean, we will define those keys in devfile/api.  and I will create another PR  to update to get from devfile/api

pkg/devfile/parser/parse.go

maysunfaisal (approved):

PR #82, opened 2021-04-23
Consume top level variables & attributes

maysunfaisal:

### What does this PR do?
- Calls the devfile/api variable func to replace the top level variable key if present in the devfile
- New interface func to get workspace content and the top-level attributes

Fixes https://github.com/devfile/api/issues/351

New go tests


pkg/devfile/parser/data/v2/attributes.go
```diff
@@ -0,0 +1,48 @@
...
+	switch d.SchemaVersion {
+	case "2.0.0":
+		return attributes.Attributes{}, fmt.Errorf("top-level attributes is not supported in devfile schema version 2.0.0")
+	default:
+		return d.Attributes, nil
+	}
+}
+
+// UpdateAttributes updates the devfile top level attributes
+func (d *DevfileV2) UpdateAttributes(attr attributes.Attributes) error {
```
yangcao77, line 21: so update attributes is going to replace the entire top-level attribute?
or should just we update for a specific key?

pkg/devfile/parser/data/v2/attributes_test.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.devfilev2.UpdateAttributes(tt.attributes)
+			if tt.wantErr == (err == nil) {
+				t.Errorf("TestUpdateAttributes error - %v, wantErr %v", err, tt.wantErr)
+			} else if err == nil {
+				attributes, err := tt.devfilev2.GetAttributes()
+				if err != nil {
+					t.Errorf("TestUpdateAttributes error2 - %+v", err)
```
yangcao77, line 107: return after this error

pkg/devfile/parser/data/v2/attributes_test.go
```diff
@@ -0,0 +1,197 @@
...
+						},
+					},
+				},
+			},
+			key:            "key3",
+			value:          nestedValue,
+			wantAttributes: attributes.Attributes{}.PutString("key1", "value1").Put("key3", nestedValue, nil).PutString("key2", "value2"),
+		},
+		{
+			name: "If Schema 2.1.0 has an attribute already present, it should overwrite",
```
yangcao77, line 163: I'm not sure if this is an expected behavior for `add`
if you check for other update & add functions for commands/components/projects etc., this should be an update behavior. for "add", if the key exist, should return an error?

yangcao77 (changes requested):

pkg/devfile/parser/data/v2/attributes.go
maysunfaisal: ok,  I will look for the specfic key for update

pkg/devfile/parser/data/v2/attributes_test.go
maysunfaisal: Do these modifications:
update will look for a specific key, err out if absent
add will just add a new key, since attributes is a map if there is a key already present, it will just overwrite with the new value

pkg/devfile/parser/data/v2/attributes.go
```diff
@@ -17,32 +17,36 @@ func (d *DevfileV2) GetAttributes() (attributes.Attributes, error) {
 	}
 }
 
-// UpdateAttributes updates the devfile top level attributes
-func (d *DevfileV2) UpdateAttributes(attr attributes.Attributes) error {
+// UpdateAttributes updates the devfile top level attribute for the specific key, err out if key is absent
+func (d *DevfileV2) UpdateAttributes(key string, value interface{}) error {
+	var err error
```
yangcao77, line 22: any reason define an err here?

pkg/devfile/parser/data/v2/attributes.go
```diff
@@ -17,32 +17,36 @@ func (d *DevfileV2) GetAttributes() (attributes.Attributes, error) {
...
-		d.Attributes = attr
+		if d.Attributes.Exists(key) {
+			d.Attributes.Put(key, value, &err)
+		} else {
+			return fmt.Errorf("cannot update top-level attribute, key %s is not present", key)
+		}
 	}
 
-	return nil
+	return err
```
yangcao77, line 36: no need for err. just return nil here or in the if statement. Same for the add function

yangcao77 (changes requested):

pkg/devfile/parser/data/v2/attributes.go
maysunfaisal: we are using it here, if i did return nil there would be one more `if err != nil` check which is the last execution in that if block. So return err is the same thing, pls check here https://github.com/devfile/library/pull/82/commits/28cfac00e0399b7fdf10700a3e9a26e3bd6d01c6#diff-ab03248fb97b24c81ae668b35d88bb6f87073ee6739cf4b6dd754d86fb685b32R30

pkg/devfile/parser/data/v2/attributes.go
maysunfaisal: because we use it here https://github.com/devfile/library/pull/82/commits/28cfac00e0399b7fdf10700a3e9a26e3bd6d01c6#diff-ab03248fb97b24c81ae668b35d88bb6f87073ee6739cf4b6dd754d86fb685b32R30

yangcao77 (approved):

PR #83, opened 2021-04-30
return error in update apis if the particular element does not exist

yangcao77:

### What does this PR do?
return error in update apis (command, component, project, starter project) if the element requested to be updated does not exist.
event update does not fall into this rule

Fixes https://github.com/devfile/api/issues/448

added & updated unit test for the error cases

### DIFF
--- a/pkg/devfile/parser/data/v2/commands.go
+++ b/pkg/devfile/parser/data/v2/commands.go
@@ -66,19 +66,13 @@ func (d *DevfileV2) AddCommands(commands []v1.Command) error {
 // UpdateCommand updates the command with the given id
 // return an error if the command is not found
 func (d *DevfileV2) UpdateCommand(command v1.Command) error {
-	found := false
 	for i := range d.Commands {
 		if d.Commands[i].Id == command.Id {
-			found = true
 			d.Commands[i] = command
-			d.Commands[i].Id = d.Commands[i].Id
-			break
+			return nil
 		}
 	}
-	if !found {
-		return fmt.Errorf("update command failed: command %s not found", command.Id)
-	}
-	return nil
+	return fmt.Errorf("update command failed: command %s not found", command.Id)
 }
 
 // DeleteCommand removes the specified command
--- a/pkg/devfile/parser/data/v2/commands.go
+++ b/pkg/devfile/parser/data/v2/commands.go
@@ -1,11 +1,10 @@
 package v2
 
 import (
-	"reflect"
-	"strings"
-
+	"fmt"
 	v1 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/library/pkg/devfile/parser/data/v2/common"
+	"reflect"
 )
 
 // GetCommands returns the slice of Command objects parsed from the Devfile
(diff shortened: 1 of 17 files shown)


pkg/devfile/parser/data/v2/commands.go
```diff
@@ -65,13 +64,21 @@ func (d *DevfileV2) AddCommands(commands []v1.Command) error {
...
+// return an error if the command is not found
+func (d *DevfileV2) UpdateCommand(command v1.Command) error {
+	found := false
 	for i := range d.Commands {
-		if strings.ToLower(d.Commands[i].Id) == strings.ToLower(command.Id) {
+		if d.Commands[i].Id == command.Id {
+			found = true
 			d.Commands[i] = command
-			d.Commands[i].Id = strings.ToLower(d.Commands[i].Id)
+			d.Commands[i].Id = d.Commands[i].Id
```
maysunfaisal, line 74: this line can be deleted

pkg/devfile/parser/data/v2/commands.go
```diff
@@ -65,13 +64,21 @@ func (d *DevfileV2) AddCommands(commands []v1.Command) error {
...
+			break
```
maysunfaisal, line 75: we can return here if we found the command and updated it

pkg/devfile/parser/data/v2/commands.go
```diff
@@ -65,13 +64,21 @@ func (d *DevfileV2) AddCommands(commands []v1.Command) error {
...
 		}
 	}
+	if !found {
+		return fmt.Errorf("update command failed: command %s not found", command.Id)
+	}
+	return nil
```
maysunfaisal, line 81: if we return instead of break on a successful update, we can remove the flag and this check and return an err at the end, we can make the same updates for the rest of the other update funcs

pkg/devfile/parser/data/v2/commands_test.go
```diff
@@ -355,7 +379,15 @@ func TestDevfile200_UpdateCommands(t *testing.T) {
...
-			d.UpdateCommand(tt.newCommand)
+			err := d.UpdateCommand(tt.newCommand)
+			// Unexpected error
+			if (err != nil) != tt.wantErr {
+				t.Errorf("TestDevfile200_UpdateCommands() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if err != nil {
+				return
+			}
```
maysunfaisal, line 390: I prefer if we put the rest of the content in a `else if err == nil {}`

if you agree, we can also do it for the other tests in this PR

pkg/devfile/parser/data/v2/projects_test.go
```diff
@@ -292,9 +293,10 @@ func TestDevfile200_UpdateProject(t *testing.T) {
 					},
 				},
 			},
+			wantErr: false,
 		},
 		{
-			name: "case:2 It should not update project for non existing project",
+			name: "It should not fail to update project for non existing project",
```
maysunfaisal, line 299: should fail*

remove the not

pkg/devfile/parser/data/v2/commands_test.go
yangcao77: updated

pkg/devfile/parser/data/v2/projects_test.go
yangcao77: updated

pkg/devfile/parser/data/v2/commands.go
yangcao77: updated

pkg/devfile/parser/data/v2/commands.go
yangcao77: updated

pkg/devfile/parser/data/v2/commands.go
yangcao77: updated

maysunfaisal (approved):

maysunfaisal:
@yangcao77 we will need to cut a new alpha release after this merges!

PR #86, opened 2021-05-10
update add apis - process everything before return an error

yangcao77:

### What does this PR do?
This PR updates add apis to process everything before return an error. 
The error returned is a total error which contains error of all invalid elements during that add action

Fixes https://github.com/devfile/api/issues/449

update unit tests to check for error message

### DIFF
--- a/pkg/devfile/parser/data/v2/commands.go
+++ b/pkg/devfile/parser/data/v2/commands.go
@@ -50,7 +50,8 @@ func (d *DevfileV2) GetCommands(options common.DevfileOptions) ([]v1.Command, er
 }
 
 // AddCommands adds the slice of Command objects to the Devfile's commands
-// if a command is already defined, error out
+// a command is considered as invalid if it is already defined
+// command list passed in will be all processed, and returns a total error of all invalid commands
 func (d *DevfileV2) AddCommands(commands []v1.Command) error {
 	var errorsList []string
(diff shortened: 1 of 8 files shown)


pkg/devfile/parser/data/v2/commands.go
```diff
@@ -51,15 +52,19 @@ func (d *DevfileV2) GetCommands(options common.DevfileOptions) ([]v1.Command, er
...
 // if a command is already defined, error out
 func (d *DevfileV2) AddCommands(commands []v1.Command) error {
-
+	var errorsList []string
 	for _, command := range commands {
 		for _, devfileCommand := range d.Commands {
 			if command.Id == devfileCommand.Id {
-				return &common.FieldAlreadyExistError{Name: command.Id, Field: "command"}
+				errorsList = append(errorsList, (&common.FieldAlreadyExistError{Name: command.Id, Field: "command"}).Error())
+				continue
```
maysunfaisal, line 60: IMO we should update teh test to see if all these ids are being returned in the err. Previously we didnt care as it was just return on first error. But since we are updating it for a specific purpose we should probably test it. We can use regex to match err strings like we do in devfile/api 🤔

yangcao77: updated

maysunfaisal (approved):

PR #89, opened 2021-05-27
add GenerateInitContainers function to generator

yangcao77:

### What does this PR do?
This PR adopts `GenerateInitContainers` and it's supporting functions `getCommandsFromEvent` and `getCommandsMap` from odo, including the unit tests associated with those functions

https://github.com/devfile/api/issues/181

add unit tests


pkg/devfile/generator/generators.go
```diff
@@ -91,6 +98,70 @@ func GetContainers(devfileObj parser.DevfileObj, options common.DevfileOptions)
...
+
+		commandsMap := getCommandsMap(commands)
+
+		for _, event := range preStartEvents {
+			eventSubCommands := getCommandsFromEvent(commandsMap, strings.ToLower(event))
+			eventCommands = append(eventCommands, eventSubCommands...)
+		}
+
+		for i, commandName := range eventCommands {
+			if command, ok := commandsMap[commandName]; ok {
```
maysunfaisal, line 124: if you think about it, we no longer have to do this check because of devfile/api validation where we check if the events are valid but i guess no harm in having it here too but removing it would remove the usage of commandsMap which is less code 🤔

pkg/devfile/generator/generators.go
```diff
@@ -91,6 +98,70 @@ func GetContainers(devfileObj parser.DevfileObj, options common.DevfileOptions)
...
+				component := common.GetExecComponent(command)
+				commandLine := common.GetExecCommandLine(command)
+				workingDir := common.GetExecWorkingDir(command)
```
maysunfaisal, line 127: so i think this is now invalid which is why odo [disabled](https://github.com/openshift/odo/blob/main/pkg/devfile/validate/events.go#L5) it as per my comment prev because `preStart` needs to be `apply`/`composite`(of only `apply` subcommands) https://github.com/devfile/api/blob/main/pkg/validation/validation-rule.md#events

pkg/devfile/generator/generators.go
```diff
@@ -91,6 +98,70 @@ func GetContainers(devfileObj parser.DevfileObj, options common.DevfileOptions)
...
+
+				var cmdArr []string
+				if workingDir != "" {
+					// since we are using /bin/sh -c, the command needs to be within a single double quote instance, for example "cd /tmp && pwd"
+					cmdArr = []string{shellExecutable, "-c", "cd " + workingDir + " && " + commandLine}
+				} else {
+					cmdArr = []string{shellExecutable, "-c", commandLine}
+				}
```
maysunfaisal, line 135: we can no longer do this because of `apply`

pkg/devfile/generator/generators.go
```diff
@@ -91,6 +98,70 @@ func GetContainers(devfileObj parser.DevfileObj, options common.DevfileOptions)
...
+
+				// Get the container info for the given component
+				for _, container := range containers {
+					if container.Name == component {
+						// override any container command and args with our event command cmdArr
+						container.Command = cmdArr
+						container.Args = []string{}
```
maysunfaisal, line 142: we should also no longer do this, and use whatever default command and args are present for a `container` component or the image defaults. So I am thinking these can be straight up be removed

maysunfaisal (changes requested):
preStart is now apply, so we should update it. Its one of the reasons why odo disabled preStart because they dont support it atm

we should also probably update odo code and remove where they call `initContainers, err := utils.GetPreStartInitContainers(a.Devfile, containers)` since its never executed 🤔

pkg/devfile/generator/generators.go
yangcao77: It's not a validation check, it's finding the particular event(command) using the event(command)'s name. If not saving it into a map, we need to iterate through the command list evenytime

maysunfaisal: oops

pkg/devfile/generator/generators.go
```diff
@@ -91,6 +98,70 @@ func GetContainers(devfileObj parser.DevfileObj, options common.DevfileOptions)
 	return containers, nil
 }
 
+// GetInitContainers gets the init container for every preStart devfile event
+func GetInitContainers(devfileObj parser.DevfileObj) ([]corev1.Container, error) {
```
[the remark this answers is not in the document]
maysunfaisal: No need to call with options, events maybe sequential/parallel. In case they are sequential, event2 may depend on event1 so would need to return all the events irrespective of the components being used by the tools or not with a custom option

pkg/devfile/generator/utils.go
```diff
@@ -451,3 +451,31 @@ func addVolumeMountToContainers(containers []corev1.Container, volumeName string
...
+		} else {
+			commands = append(commands, command.Id)
+		}
+	}
+
+	return commands
+}
+
+// getCommandsMap returns a map of the command Id to the command
+func getCommandsMap(commands []v1.Command) map[string]v1.Command {
```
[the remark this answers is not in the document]
yangcao77: moved to `command_helper` under `/pkg/devfile/parser/data/v2/common`

pkg/devfile/generator/utils.go
```diff
@@ -451,3 +451,31 @@ func addVolumeMountToContainers(containers []corev1.Container, volumeName string
 		}
 	}
 }
+
+// getCommandsFromEvent returns the list of commands from the event name.
+// If the event is a composite command, it returns the sub-commands from the tree
+func getCommandsFromEvent(commandsMap map[string]v1.Command, eventName string) []string {
```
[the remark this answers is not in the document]
yangcao77: moved to `command_helper` under `/pkg/devfile/parser/data/v2/common`

pkg/devfile/generator/generators.go
yangcao77: updated

pkg/devfile/generator/generators.go
yangcao77: updated

pkg/devfile/generator/generators.go
yangcao77: updated

pkg/devfile/parser/data/v2/common/command_helper.go
```diff
@@ -50,6 +50,33 @@ func GetExecWorkingDir(dc v1.Command) string {
...
+
+	return ""
+}
+
+//// GetApplyCommandLine returns the command line of the exec command
+//func GetApplyCommandLine(dc v1.Command) string {
+//	if dc.Apply != nil {
+//		return dc.Apply
+//	}
+//
```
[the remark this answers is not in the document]
yangcao77: removed

pkg/devfile/parser/data/v2/common/command_helper.go
```diff
@@ -50,6 +50,33 @@ func GetExecWorkingDir(dc v1.Command) string {
 	return ""
 }
 
+// GetApplyComponent returns the component of the exec command
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/devfile/parser/data/v2/common/command_helper.go
```diff
@@ -66,3 +93,31 @@ func GetCommandType(command v1.Command) (v1.CommandType, error) {
 		return "", fmt.Errorf("unknown command type")
 	}
 }
+
+// GetCommandsMap returns a map of the command Id to the command
+func GetCommandsMap(commands []v1.Command) map[string]v1.Command {
```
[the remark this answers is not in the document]
yangcao77: This is a fairly simple function, so we can skip the unit test for it.

maysunfaisal (approved):

PR #93, opened 2021-06-04
Add mock func for DevfileData interface

maysunfaisal:

### What does this PR do?
Adds mock funcs for the DevfileData interface via the mockgen tool

Fixes https://github.com/devfile/api/issues/383

Yes, updated tests to use the mock data


pkg/devfile/generator/generators_test.go
```diff
@@ -364,33 +386,32 @@ func TestGetVolumesAndVolumeMounts(t *testing.T) {
...
 
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+			mockDevfileData := data.NewMockDevfileData(ctrl)
+
+			// set up the mock data
+			if tt.wantErr {
+				// if we have an error, just call the mock GetDevfileContainerComponents once for GetContainers() and a different
+				// one for GetVolumesAndVolumeMounts() below to simulate err from that function
+				mockDevfileData.EXPECT().GetDevfileContainerComponents(common.DevfileOptions{}).Return(tt.components, nil).Times(1)
```
yangcao77, line 406: do we have to declare times at this place?
I'm not familiar with gomock, but I'm thinking if it's possible to 
```
mockDevfileData.EXPECT().GetDevfileContainerComponents(common.DevfileOptions{}).Return(tt.components, nil)
...
GetContainers()
if tt.wantErr {
  // simulate error condition
  mockDevfileData.EXPECT().GetDevfileContainerComponents(common.DevfileOptions{}).Return(nil, fmt.Errorf("mock error")).Times(1)
}
GetVolumesAndVolumeMounts()
```

maysunfaisal: I tried it out, it does not work. Its treated as one call. I tried out MinTimes(1) as well but that too doesnt seem to work. For without anything:
```
= RUN   TestGetVolumesAndVolumeMounts/One_volume_mounted
    generators.go:360: Unexpected call to *data.MockDevfileData.GetDevfileContainerComponents([{map[] { } {} {}}]) at /Users/maysun/dev/devfile/parser/pkg/devfile/generator/generators.go:360 because: 
        expected call at /Users/maysun/dev/devfile/parser/pkg/devfile/generator/generators_test.go:411 has already been called the max number of times
```

yangcao77: everytime you declare the `mockDevfileData.EXPECT().GetDevfileContainerComponents(common.DevfileOptions{})` is creating a new `*Call`.

I have tested the following works: which no need for the first if-else statement to set different call times, and only need to re-declare the return value for error case. this simplifies the code. also suggest to change for other applicable tests
```suggestion
			getContainerCall := mockDevfileData.EXPECT().GetDevfileContainerComponents(common.DevfileOptions{})
			// set up the mock data
			getContainerCall.Return(tt.components, nil).AnyTimes()
			mockDevfileData.EXPECT().GetProjects(common.DevfileOptions{}).Return(nil, nil).AnyTimes()
			devObj := parser.DevfileObj{
				Data: mockDevfileData,
			}

			containers, err := GetContainers(devObj, common.DevfileOptions{})
			if err != nil {
				t.Errorf("TestGetVolumesAndVolumeMounts error - %v", err)
				return
			}

			if tt.wantErr {
				// simulate error condition
				getContainerCall.Return(nil, fmt.Errorf("mock error"))
			}
```

yangcao77:
Should the mock_interface be under testingutil pkg, instead of our production pkg?

maysunfaisal:
@yangcao77 its usually kept in the same dir as the interface file. Thats the practice that I saw in online examples and odo as our close reference.

yangcao77:
> @yangcao77 its usually kept in the same dir as the interface file. Thats the practice that I saw in online examples and odo as our close reference.
> 
> I'll check more tho, brb

Reasonable. If we have new interface for v3 in the future, we need to create matching mock funcs for new interface anyways.

pkg/devfile/generator/generators_test.go
maysunfaisal: nice

yangcao77:
Another thought on this PR:
Since `GetDevfileContainerComponents` and `GetDevfileVolumeComponents` are depracated functions, our library should switch to use `GetComponents(common.DevfileOptions)` instead

pkg/devfile/generator/generators_test.go
maysunfaisal: updated

yangcao77 (approved):
changes look good to me

PR #94, opened 2021-06-07
run action to get go pkg upon new release

yangcao77:
### What does this PR do?

this pr runs an action to `get devfile/library` upon new release. so that the latest release can get updated in pkg.go.dev automatically. 

Fixes https://github.com/devfile/api/issues/486


.github/workflows/proxy-warming.yml
```diff
@@ -0,0 +1,17 @@
+name: Renew documentation
+
+on:
+  release:
+    types:
+      - created
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - '**/v[0-9]+.[0-9]+.[0-9]+'
```
maysunfaisal, line 9: nice, but this will not do it for the future alpha or beta releases 🤔 

or any releases with letters in them

maysunfaisal (approved):

PR #102, opened 2021-06-23
Add events and metadata tests

kim-tsao:
### What does this PR do?
* Adds tests for events and metadata objects.  
* Bug fix in the devfileFollower.LibraryData.UpdateEvents call, where parameters were out of order
* Removed unused TestContent struct in library_test.go.  We are using the one from the common utils

https://github.com/devfile/api/issues/502

This was tested locally.  Output in generated yamls and test.log were verified


tests/v2/libraryTest/library_test.go
```diff
@@ -8,17 +8,10 @@ import (
...
-	CommandTypes   []schema.CommandType
-	ComponentTypes []schema.ComponentType
-	FileName       string
-	EditContent    bool
-}
-
 func Test_ExecCommand(t *testing.T) {
 	testContent := commonUtils.TestContent{}
 	testContent.CommandTypes = []schema.CommandType{schema.ExecCommandType}
+	testContent.AddEvents = commonUtils.GetBinaryDecision()
```
yangcao77, line 14: `AddEvents ` randomly set event values: https://github.com/devfile/api/blob/b08e48618403f7102592389e621531145bdfcbcc/test/v200/utils/common/events-utils.go#L35
looks like you cannot control to be only `Exec` or `Apply` or `composite` command.
Should not be under test for each particular type of command.

kim-tsao: ok, I see what you mean.  SetEventValues will create extra commands that will be outside the scope of the Exec testing, so I need to narrow this down to just the commands that Exec or composite Exec supports

kim-tsao: I think the easiest thing to do in this case is what you suggested; to remove events testing in these scenarios (otherwise I'd have to introduce new functions in the api utils).  I added new events test cases similar to what's being done in api tests, so that should ensure sufficient coverage.

tests/v2/libraryTest/library_test.go
```diff
@@ -73,6 +65,7 @@ func Test_MultiCommand(t *testing.T) {
 	testContent.CommandTypes = []schema.CommandType{schema.ExecCommandType,
 		schema.CompositeCommandType,
 		schema.ApplyCommandType}
+	testContent.AddEvents = commonUtils.GetBinaryDecision()
```
yangcao77, line 68: should this be removed?

kim-tsao: yes, looks like I missed that one

yangcao77 (approved):
lgtm

PR #115, opened 2021-09-15
update tests and utils to use boolean pointers

kim-tsao:
### What does this PR do?:

- Updates tests and utils to use boolean pointers
- Updates the parser tests to v2.1.0 and expands coverage to include all boolean fields

https://github.com/devfile/api/issues/545

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [X ] Unit/Functional tests

  

N/A
- [ ] Documentation 

   
N/A

- [ ] Client Impact

  
Client impact discussion in issue https://github.com/devfile/api/issues/545

### DIFF
--- a/pkg/devfile/generator/utils_test.go
+++ b/pkg/devfile/generator/utils_test.go
@@ -22,6 +22,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+var isTrue bool = true
+
 func TestConvertEnvs(t *testing.T) {
 	envVarsNames := []string{"test", "sample-var", "myvar"}
 	envVarsValues := []string{"value1", "value2", "value3"}
@@ -984,7 +986,7 @@ func TestGetPortExposure(t *testing.T) {
 								{
 									Name:       urlName2,
 									TargetPort: 9090,
-									Secure:     true,
+									Secure:     &isTrue,
 									Path:       "/testpath",
 									Exposure:   v1.InternalEndpointExposure,
 									Protocol:   v1.HTTPSEndpointProtocol,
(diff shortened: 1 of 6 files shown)


pkg/devfile/generator/utils_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+var isTrue bool = true
```
yangcao77, line 25: suggest to use `const` instead of `var`. Same for all other `isTrue`/`isFalse`
```suggestion
const isTrue bool = true
```

pkg/devfile/parser/parse_test.go
```diff
@@ -41,7 +45,7 @@ func Test_parseParentAndPluginFromURI(t *testing.T) {
 		Data: &v2.DevfileV2{
 			Devfile: v1.Devfile{
 				DevfileHeader: devfilepkg.DevfileHeader{
-					SchemaVersion: schemaV200,
+					SchemaVersion: schemaV210,
```
yangcao77, line 48: should be `v2.2.0`, the bool pointer change is not in v2.1.0. 
we should also update all other tests to test with latest schema. probably create a file for const values will be used in tests, including latest_schema, isTrue and isFalse, and etc. 
 it doesn't have to be within this PR, we can create another issue to track that.

yangcao77 (changes requested):

pkg/devfile/generator/utils_test.go
kim-tsao: Constants are not addressable types: https://golang.org/ref/spec#Address_operators so this will cause a compile error when assigning the address of the constant

pkg/devfile/parser/parse_test.go
kim-tsao: The bool pointer updates do not affect the schema structure, just the API.   I needed to bump up the schema versions so we could test the `ephemeral` property which did not exist in v2.0.0.  We can consider refactoring the bool variables and possibly other utils to a common file for all tests (unit, functional) to reference.

yangcao77 (approved):

PR #122, opened 2021-10-27
Set default boolean values

kim-tsao:
### What does this PR do?:

- Updates the parser to set defaults on nil boolean fields after workspace contents are merged.

- Updates the parser tests to include coverage for overriding unset/set values, and merging values for all boolean properties including the `RootRequired` field in the 2.2 spec. 

https://github.com/devfile/api/issues/615

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [ X] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  
These changes ensure clients do not encounter nil boolean values when using a parsed object, similar to the expected content before boolean pointers were introduced.

### DIFF
--- a/pkg/devfile/parser/parse.go
+++ b/pkg/devfile/parser/parse.go
@@ -288,7 +288,6 @@ func parseParentAndPlugin(d DevfileObj, resolveCtx *resolutionContextTree, tool 
 	d.Data.SetDevfileWorkspaceSpecContent(*mergedContent)
 	// remove parent from flatterned devfile
 	d.Data.SetParent(nil)
-	//setDefaults(d)
 
 	return nil
 }
--- a/pkg/devfile/parser/parse.go
+++ b/pkg/devfile/parser/parse.go
@@ -281,6 +281,7 @@ func parseParentAndPlugin(d DevfileObj, resolveCtx *resolutionContextTree, tool 
 	d.Data.SetDevfileWorkspaceSpecContent(*mergedContent)
 	// remove parent from flatterned devfile
 	d.Data.SetParent(nil)
+	setDefaults(d)
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


pkg/devfile/parser/parse.go
```diff
@@ -281,6 +288,7 @@ func parseParentAndPlugin(d DevfileObj, resolveCtx *resolutionContextTree, tool
 	d.Data.SetDevfileWorkspaceSpecContent(*mergedContent)
 	// remove parent from flatterned devfile
 	d.Data.SetParent(nil)
+	//setDefaults(d)
```
yangcao77, line 291: should remove this comment?

pkg/devfile/parser/parse_test.go
```diff
@@ -133,6 +135,27 @@ func Test_parseParentAndPluginFromURI(t *testing.T) {
 									},
 								},
 							},
+							{
+								Name: "image",
```
yangcao77, line 139: those image components are basically identical, except for those boolean values. Can you create a function to generate the image component and accept the boolean value as a param? and call that func directly instead of re-defining it in each test

yangcao77 (changes requested):

pkg/devfile/parser/parse_test.go
kim-tsao: good point, optimizations can be made in other components too.  I'll try to get those updated as well.

yangcao77: There are some common functions for other components in: https://github.com/devfile/library/blob/main/pkg/testingutil/devfile.go  , you can add image component in that file as well

for those tests that need more flexibility on component content can stay with the definition in unit test.

kim-tsao: I created some utils and I was able to extend it so other properties could be set.  If you agree with this, I think we should look  at cleaning up some areas of our test cases in a similar manner since it produces a more readable file

yangcao77 (approved):
Changes lgtm

PR #128, opened 2022-01-17
use endpoint name as container port name and service port name

yangcao77:

### What does this PR do?:

use endpoint name as container port name and service port name if <= 15 chars. if > 15 chars (devfile spec before the change to shorten endpoint name limit), use `port-<portNumber>`

Fixes https://github.com/devfile/api/issues/700

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/pkg/devfile/generator/utils.go
+++ b/pkg/devfile/generator/utils.go
@@ -46,7 +46,12 @@ func convertPorts(endpoints []v1.Endpoint) []corev1.ContainerPort {
 		} else {
 			portProtocol = corev1.ProtocolTCP
 		}
-		name := fmt.Sprintf("%d-%s", portNumber, strings.ToLower(string(portProtocol)))
+		name := endpoint.Name
+		if len(name) > 15 {
+			// to be compatible with endpoint longer than 15 chars
+			name = fmt.Sprintf("port-%v", portNumber)
+		}
+
 		if _, exist := portMap[name]; !exist {
 			portMap[name] = true
 			containerPorts = append(containerPorts, corev1.ContainerPort{
@@ -268,7 +273,6 @@ func getServiceSpec(devfileObj parser.DevfileObj, selectorLabels map[string]stri
 			}
 			// if Exposure == none, should not create a service for that port
 			if !portExist && portExposureMap[int(port.ContainerPort)] != v1.NoneEndpointExposure {
-				port.Name = fmt.Sprintf("port-%v", port.ContainerPort)
 				containerPorts = append(containerPorts, port)
 			}
 		}
--- a/pkg/devfile/generator/utils_test.go
+++ b/pkg/devfile/generator/utils_test.go
@@ -97,7 +97,7 @@ func TestConvertEnvs(t *testing.T) {
 }
 
 func TestConvertPorts(t *testing.T) {
-	endpointsNames := []string{"endpoint1", "endpoint2"}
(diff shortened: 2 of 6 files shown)


yangcao77:
@valaparthvi can you also help verify this PR fixes the Odo issue?

valaparthvi:
@yangcao77 It works like magic! Thank you for fixing it.
So far I've tested with the following cases, and all seem to work as expected.
1. passing custom names as mentioned in https://github.com/redhat-developer/odo/issues/4737
2. without passing any name
3. with duplicate endpoints
4. passing a very long name
I hope I didn't miss out on any case.

pkg/devfile/parser/data/v2/2.2.0/devfileJsonSchema220.go
```diff
@@ -347,7 +347,7 @@ const JsonSchema220 = `{
                     },
                     "name": {
                       "type": "string",
-                      "maxLength": 63,
+                      "maxLength": 15,
```
kim-tsao, line 350: do we have corresponding updates in our API docs to show the new max length?

pkg/devfile/generator/utils_test.go
```diff
@@ -663,7 +679,7 @@ func TestGetServiceSpec(t *testing.T) {
 		wantPorts           []corev1.ServicePort
 	}{
 		{
-			name: "multiple endpoints share the same port",
+			name: "multiple endpoints have different ports",
```
kim-tsao, line 682: This test should cover most of our scenarios since we have validation on unique ports but should we handle the exception case as well where `dedicatedPod=true`? https://github.com/devfile/api/blob/main/pkg/validation/validation-rule.md#endpoints

pkg/devfile/generator/utils.go
```diff
@@ -46,7 +46,12 @@ func convertPorts(endpoints []v1.Endpoint) []corev1.ContainerPort {
 		} else {
 			portProtocol = corev1.ProtocolTCP
 		}
-		name := fmt.Sprintf("%d-%s", portNumber, strings.ToLower(string(portProtocol)))
+		name := endpoint.Name
```
kim-tsao, line 49: We are changing a user provided name underneath the covers if length exceeds 15 characters.  It'll be good to warn them of these changes ahead of time (via cmd line, log, and/or doc)

pkg/devfile/parser/data/v2/2.2.0/devfileJsonSchema220.go
yangcao77: the spec is the api doc

pkg/devfile/generator/utils_test.go
yangcao77: we do not specifially do tests for `deficatedPod` in library, since it is currently not supported in library. see https://github.com/devfile/api/issues/670

pkg/devfile/generator/utils.go
yangcao77: yes. This change is because we want to unify the service port name & the container port name. since service port does not accept starting with a number (has to be alphabet), thus we use the convention `port-<number>`
@valaparthvi  FYI.

kim-tsao: Does the warning need to happen on the `odo` side?

yangcao77: correct

kim-tsao (approved):

PR #129, opened 2022-01-25
Add Replicas field to DeploymentParams struct

dharmit:
### What does this PR do?:

Adds `Replicas` field to `DeploymentParams` struct. This will help initialize a Deployment with Replica set to what the tooling would like to have. By default, library doesn't set it to anything at the moment.

https://github.com/redhat-developer/odo/issues/5347

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/pkg/devfile/generator/generators.go
+++ b/pkg/devfile/generator/generators.go
@@ -2,8 +2,6 @@ package generator
 
 import (
 	"fmt"
-	"k8s.io/utils/pointer"
-
 	v1 "github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	"github.com/devfile/library/pkg/devfile/parser"
 	"github.com/devfile/library/pkg/devfile/parser/data/v2/common"
@@ -159,7 +157,7 @@ type DeploymentParams struct {
 	Containers        []corev1.Container
 	Volumes           []corev1.Volume
 	PodSelectorLabels map[string]string
-	Replicas          int32
+	Replicas          *int32
 }
 
 // GetDeployment gets a deployment object
@@ -189,8 +187,8 @@ func GetDeployment(devfileObj parser.DevfileObj, deployParams DeploymentParams) 
 		Spec:       *getDeploymentSpec(deploySpecParams),
 	}
 
-	if deployParams.Replicas != 0 {
-		deployment.Spec.Replicas = pointer.Int32Ptr(deployParams.Replicas)
+	if deployParams.Replicas != nil {
+		deployment.Spec.Replicas = deployParams.Replicas
 	}
 
 	return deployment, nil
--- a/pkg/devfile/generator/generators.go
+++ b/pkg/devfile/generator/generators.go
@@ -2,6 +2,7 @@ package generator
 
 import (
 	"fmt"
+	"k8s.io/utils/pointer"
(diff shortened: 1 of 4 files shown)


pkg/devfile/generator/generators.go
```diff
@@ -158,6 +159,7 @@ type DeploymentParams struct {
 	Containers        []corev1.Container
 	Volumes           []corev1.Volume
 	PodSelectorLabels map[string]string
+	Replicas          int32
```
feloy, line 162: I would set the Replicas as a pointer, to make it optional (the nil value indicating the user does not want to set it). The 0 value on which you test below is a valid value a user would want to set.

dharmit: Makes sense. Does https://github.com/devfile/library/pull/129/commits/145564efc02c37f573608a4875e81009f3da3427 look good?

pkg/devfile/generator/generators.go
```diff
@@ -187,6 +187,10 @@ func GetDeployment(devfileObj parser.DevfileObj, deployParams DeploymentParams)
 		Spec:       *getDeploymentSpec(deploySpecParams),
 	}
 
+	if deployParams.Replicas != nil {
+		deployment.Spec.Replicas = deployParams.Replicas
```
feloy, line 191: I think that this affectation should be done in getDeploymentSpec which is the dedicated function for this.

dharmit: I agree. That's where I first looked. But that would mean changing the signature of `getDeploymentSpec`. I was not sure if that would be OK. But it's an internal function, so changing signature shouldn't be a big deal.

Pushed https://github.com/devfile/library/pull/129/commits/4a544b12a31ed89f58f1da0e08fa074ab203726a. PTAL.

pkg/devfile/generator/generators.go
```diff
@@ -184,7 +184,7 @@ func GetDeployment(devfileObj parser.DevfileObj, deployParams DeploymentParams)
 	deployment := &appsv1.Deployment{
 		TypeMeta:   deployParams.TypeMeta,
 		ObjectMeta: deployParams.ObjectMeta,
-		Spec:       *getDeploymentSpec(deploySpecParams),
+		Spec:       *getDeploymentSpec(deploySpecParams, deployParams),
```
yangcao77, line 187: add it to be an additional field in `deploySpecParams` instead of changing the function input

yangcao77 (changes requested):

pkg/devfile/generator/generators.go
dharmit: @yangcao77 thanks, I have made the change.

@yangcao77 @feloy What surprised me was I no longer needed the `if` check. I was expecting a nil pointer error, but it seems that was my lack of understanding of Go.

dharmit:
@yangcao77 do I need to make any changes to the PR? Would `go mod tidy` suffice?

yangcao77:
@dharmit  yes, please run `go mod tidy` and `go fmt -x ./...`

yangcao77 (approved):

PR #137, opened 2022-08-08
Add Kube YAML parser/reader

maysunfaisal:

### What does this PR do?:
Adds in the Kubernetes YAML parser. This allows us to parse and read the YAML files specified in a Kubernetes/OpenShift component in a devfile

Three srcs - Path, URL and Data

<TBD>

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

- Call the function with a Kubernetes YAML as one of the src
- Or, alternatively run the test specified


pkg/devfile/parser/reader.go
```diff
@@ -0,0 +1,107 @@
...
+	var data []byte
+	var err error
+
+	if src.URL != "" {
+		data, err = util.DownloadFileInMemory(src.URL)
+		if err != nil {
+			return nil, nil, nil, nil, errors.Wrapf(err, "failed to download file %q", src.URL)
+		}
+	} else if src.Path != "" {
+		absPath, err := util.GetAbsPath(src.Path)
```
yangcao77, line 43: probably a dumb question... is the `src.Path` you passing in here a relative path to devfile.yaml?
so the `GetAbsPath` is calsulating from the user's current path... if the devfile.yaml is under `/user/devfilefolder/devfile.yaml` and user is running parser from `/user/`.  will the `GetAbsPath()` try to look the relative path against `/user/` instead of `/user/devfilefolder/`?

yangcao77: and.. what if the devfile.yaml we are parsing is from a URL? it cannot be resolved by `GetAbsPath`

pkg/devfile/parser/reader.go
```diff
@@ -0,0 +1,107 @@
...
+	URL string
+	// Data is the yaml content in []byte format.
+	Data []byte
+}
+
+// ReadKubernetesYaml reads a yaml Kubernetes file from either the Path, URL or Data provided.
+// It returns Deployments, Services, Routes resources as the primary Kubernetes resources.
+// Other Kubernetes resources are returned as []byte type. Consumers interested in other Kubernetes resources
+// are expected to Unmarshal it to the struct of the respective resource.
+func ReadKubernetesYaml(src YamlSrc, fs filesystem.Filesystem) ([]appsv1.Deployment, []corev1.Service, []routev1.Route, [][]byte, error) {
```
yangcao77, line 32: How are we expecting to use this function? consumer directly call?
I'm wondering should this function be called by `generator` when consumer trying to run `getDeployment`/`getService`/`getRoute`?

pkg/devfile/parser/reader.go
```diff
@@ -0,0 +1,107 @@
...
+
+	var data []byte
+	var err error
+
+	if src.URL != "" {
+		data, err = util.DownloadFileInMemory(src.URL)
+		if err != nil {
+			return nil, nil, nil, nil, errors.Wrapf(err, "failed to download file %q", src.URL)
```
yangcao77, line 40: instead of having much returning values, can create a sturct
```
type KubernetesResources struct {
	Deployments []appsv1.Deployment
	Services []corev1.Service
	Routes []routev1.Route
        OtherResources [][]byte
}
```
and then return `(KubernetesResources, error)`

pkg/devfile/parser/reader.go
maysunfaisal: good question, let me ask now in the meeting

pkg/devfile/parser/reader.go
```diff
@@ -0,0 +1,107 @@
...
+// It returns Deployments, Services, Routes resources as the primary Kubernetes resources.
+// Other Kubernetes resources are returned as []byte type. Consumers interested in other Kubernetes resources
+// are expected to Unmarshal it to the struct of the respective resource.
+func ReadKubernetesYaml(src YamlSrc, fs filesystem.Filesystem) ([]appsv1.Deployment, []corev1.Service, []routev1.Route, [][]byte, error) {
+
+	var data []byte
+	var err error
+
+	if src.URL != "" {
+		data, err = util.DownloadFileInMemory(src.URL)
```
johnmcollier, line 38: Can we separate the logic to download / read the Kubernetes yaml file, from the logic to parse the files resources? I.e. maybe modify `ReadKubernetesYaml` to take in a byte array?

tests/yamls/resources.yaml
```diff
@@ -0,0 +1,120 @@
+apiVersion: apps/v1
```
johnmcollier, line 1: Can you a test file with multiple deployments/service/other resources?

johnmcollier (commented):
Just two suggestions.

pkg/devfile/parser/reader.go
maysunfaisal: For URL there is `src.URL`; let me take a look at the AbsPath case

maysunfaisal: Updated, not getting the abs path. Leaving it up to the consumer to handle the path. Utility func job is to just read from the path/url/data provided

pkg/devfile/parser/reader.go
maysunfaisal: updated

pkg/devfile/parser/reader.go
maysunfaisal: updated

tests/yamls/resources.yaml
maysunfaisal: updated

pkg/devfile/parser/reader.go
yangcao77: discussed in the team meeting. to keep it simple, we currently only provide `ReadKubernetesYaml` as a helper function, and tool should extract the content they need from the returned result.

yangcao77 (approved):
changes look good to me

johnmcollier (approved):

PR #138, opened 2022-08-11
906- variable substitution in kubernetes uri content

yangcao77:
### What does this PR do?:

This PR converts kubernetest `uri` field to the `inlined` field with actual content in the `devfileObj` object in memory; so that the grobal variable defined can be substituted. An attribute `devfile.io/kubeComponent-originalURI` will be added to the kubernetes component as well. 
The PR also modified writer to check for the specific attribute and restore the uri back before write to the devfile; so that we are not modifying user's devfile content. 

fixes https://github.com/devfile/api/issues/906

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

### DIFF
--- a/pkg/devfile/parser/devfileobj.go
+++ b/pkg/devfile/parser/devfileobj.go
@@ -7,8 +7,8 @@ import (
 
 // Default filenames for create devfile
 const (
-	OutputDevfileYamlPath       = "devfile.yaml"
-	KubeComponentOriginalURIKey = "devfile.io/kubeComponent-originalURI"
+	OutputDevfileYamlPath          = "devfile.yaml"
+	K8sLikeComponentOriginalURIKey = "devfile.io/k8sLikeComponent-originalURI"
 )
 
 // DevfileObj is the runtime devfile object
--- a/pkg/devfile/parser/devfileobj.go
+++ b/pkg/devfile/parser/devfileobj.go
@@ -7,7 +7,7 @@ import (
 
 // Default filenames for create devfile
 const (
-	OutputDevfileYamlPath = "devfile.yaml"
+	OutputDevfileYamlPath       = "devfile.yaml"
 	KubeComponentOriginalURIKey = "devfile.io/kubeComponent-originalURI"
 )
 
--- a/pkg/devfile/parser/devfileobj.go
+++ b/pkg/devfile/parser/devfileobj.go
@@ -8,6 +8,7 @@ import (
 // Default filenames for create devfile
 const (
 	OutputDevfileYamlPath = "devfile.yaml"
+	KubeComponentOriginalURIKey = "devfile.io/kubeComponent-originalURI"
 )
 
 // DevfileObj is the runtime devfile object
--- a/pkg/devfile/parser/devfileobj.go
+++ b/pkg/devfile/parser/devfileobj.go
@@ -8,7 +8,7 @@ import (
 // Default filenames for create devfile
 const (
 	OutputDevfileYamlPath          = "devfile.yaml"
-	K8sLikeComponentOriginalURIKey = "devfile.io/k8sLikeComponent-originalURI"
+	K8sLikeComponentOriginalURIKey = "api.devfile.io/k8sLikeComponent-originalURI"
 )
 
 // DevfileObj is the runtime devfile object
--- a/pkg/devfile/parser/parse.go
+++ b/pkg/devfile/parser/parse.go
@@ -633,7 +633,7 @@ func parseKubeComponentFromURI(devObj DevfileObj) error {
 }
 
 //convertKubeCompUriToInlined read in kubernetes resources definition from uri and converts to kubernetest inlined field
-func convertKubeCompUriToInlined(component *v1.Component, d devfileCtx.DevfileCtx) error{
+func convertKubeCompUriToInlined(component *v1.Component, d devfileCtx.DevfileCtx) error {
 	uri := component.Kubernetes.Uri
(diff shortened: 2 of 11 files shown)


pkg/devfile/parser/parse.go
```diff
@@ -49,6 +50,11 @@ func parseDevfile(d DevfileObj, resolveCtx *resolutionContextTree, tool resolver
 		return d, errors.Wrapf(err, "failed to decode devfile content")
 	}
 
+	err = parseKubeComponentFromURI(d)
```
kim-tsao, line 53: does it make sense to parse after flattening in case there are parent overrides for the kubernetes uri?

pkg/devfile/parser/parse.go
```diff
@@ -599,3 +605,78 @@ func setEndpoints(endpoints []v1.Endpoint) {
...
+			ComponentType: v1.KubernetesComponentType,
+		},
+	}
+	kubeComponents, err := devObj.Data.GetComponents(getKubeCompOptions)
+	if err != nil {
+		return err
+	}
+	for _, kubeComp := range kubeComponents {
+		if kubeComp.Kubernetes.Uri != "" {
+			err := convertKubeCompUriToInlined(&kubeComp, devObj.Ctx)
```
kim-tsao, line 622: Is there a way to determine if the inlined content contains substitution variables?  If it doesn't then we don't need to process it

pkg/devfile/parser/writer.go
```diff
@@ -11,6 +14,11 @@ import (
 // WriteYamlDevfile creates a devfile.yaml file
 func (d *DevfileObj) WriteYamlDevfile() error {
 
+	// Check kubernetes components, and restore original uri content
+	err := restoreK8sCompURI(d)
```
maysunfaisal, line 18: just a thought:
this func would be checking for the attribute every time it is called to write devfile. I wonder if we should introduce some property in the object that we can use to keep track:

```

type DevfileObj struct {

	// Ctx has devfile context info
	Ctx devfileCtx.DevfileCtx

	// Data has the devfile data
	Data data.DevfileData

        // convert
        ConvertUriToInlined bool
}
```

pkg/devfile/parser/writer.go
```diff
@@ -30,3 +38,58 @@ func (d *DevfileObj) WriteYamlDevfile() error {
...
+	openshiftComponents, err := devObj.Data.GetComponents(getOpenshiftCompOptions)
+	if err != nil {
+		return err
+	}
+	for _, kubeComp := range kubeComponents {
+		var keyNotFoundErr = &apiAttributes.KeyNotFoundError{Key: K8sLikeComponentOriginalURIKey}
+		uri := kubeComp.Attributes.GetString(K8sLikeComponentOriginalURIKey, &err)
+		if err != nil && err.Error() != keyNotFoundErr.Error() {
+			return err
+		}
```
maysunfaisal, line 66: you can also do it this way:

```suggestion
		if err != nil {
		        if _, ok := err.(*attributes.KeyNotFoundError); !ok {
			        return err
		        }
	        }
```

pkg/devfile/parser/parse.go
yangcao77: my bad, uri should not be converted to inlined until flattenned. updated

pkg/devfile/parser/devfileobj.go
```diff
@@ -7,7 +7,8 @@ import (
 
 // Default filenames for create devfile
 const (
-	OutputDevfileYamlPath = "devfile.yaml"
+	OutputDevfileYamlPath          = "devfile.yaml"
+	K8sLikeComponentOriginalURIKey = "devfile.io/k8sLikeComponent-originalURI"
```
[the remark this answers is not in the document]
yangcao77: updated

pkg/devfile/parser/parse.go
```diff
@@ -599,3 +617,117 @@ func setEndpoints(endpoints []v1.Endpoint) {
...
+	kubeComponents, err := devObj.Data.GetComponents(getKubeCompOptions)
+	if err != nil {
+		return err
+	}
+	openshiftComponents, err := devObj.Data.GetComponents(getOpenshiftCompOptions)
+	if err != nil {
+		return err
+	}
+	for _, kubeComp := range kubeComponents {
+		if kubeComp.Kubernetes.Uri != "" {
```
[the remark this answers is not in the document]

pkg/devfile/parser/writer.go
yangcao77: updated

pkg/devfile/parser/writer.go
yangcao77: I do not want the value to be direct accessible.. so added into `DevfileObj.ctx.convertUriToInlined` instead.

pkg/devfile/parser/parse.go
yangcao77: it will still need to read in the content, and call `ValidateAndReplaceGlobalVariable` to look for global variable and validation. that will be too much extra work compare with convert all to inlined and only call `ValidateAndReplaceGlobalVariable` after parsing

pkg/devfile/parser/writer.go
maysunfaisal: Thinking if we should really keep this context? Because context refers to a devfile context but otherwise its ok i guess

maysunfaisal (approved):

PR #139, opened 2022-08-16
parsing devfile downloads resource files

mike-hoang:
Signed-off-by: Michael Hoang <mhoang@redhat.com>

### What does this PR do?:
Downloads all resources files from a stack to the local root folder if it doesn’t already exist (will not overwrite files).

The resource files will only be downloaded in the following cases:

```yaml
# with a valid registry url and id
parent:
  id: nodejs
  registryUrl: https://registry.devfile.io
```
```yaml
# with a raw GitHub file link that points to a devfile
parent:
  uri: https://raw.githubusercontent.com/devfile/registry/main/stacks/nodejs/devfile.yaml
```
```yaml
# with a relative path
parent:
  uri: ./parent/devfile.yaml
```

fixes devfile/api#46

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact


pkg/devfile/parser/parse.go
```diff
@@ -397,11 +400,24 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
...
+	tempDir, _ := ioutil.TempDir("", "")
+	defer os.RemoveAll(tempDir)
+
 	if registryURL != "" {
-		devfileContent, err := getDevfileFromRegistry(id, registryURL, importReference.Version)
+		err = getStackFromRegistry(id, registryURL, tempDir, destDir)
 		if err != nil {
 			return DevfileObj{}, err
 		}
+		devfileContent, err := getDevfileFromDir(tempDir)
```
yangcao77, line 414: should be `destDir`?
```suggestion
		devfileContent, err := getDevfileFromDir(destDir)
```

pkg/devfile/parser/parse.go
```diff
@@ -397,11 +400,24 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
...
 	registryURL := importReference.RegistryUrl
+	destDir := "."
+
+	// contains the downloaded stack from the registry
+	tempDir, _ := ioutil.TempDir("", "")
+	defer os.RemoveAll(tempDir)
+
 	if registryURL != "" {
-		devfileContent, err := getDevfileFromRegistry(id, registryURL, importReference.Version)
+		err = getStackFromRegistry(id, registryURL, tempDir, destDir)
```
yangcao77, line 410: to locate & download a specific version from registry, the `Version` information is also required. https://github.com/devfile/api/blob/351f05b7c2b18e8a449dad444029cdf5e1ca81cc/pkg/apis/workspaces/v1alpha2/import_reference.go#L60

pkg/devfile/parser/parse.go
```diff
@@ -397,11 +400,24 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
 func parseFromRegistry(importReference v1.ImportReference, resolveCtx *resolutionContextTree, tool resolverTools) (d DevfileObj, err error) {
 	id := importReference.Id
 	registryURL := importReference.RegistryUrl
+	destDir := "."
+
+	// contains the downloaded stack from the registry
+	tempDir, _ := ioutil.TempDir("", "")
```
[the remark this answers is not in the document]
yangcao77: might want to move the tempDir creation piece to `getStackFromRegistry`

yangcao77 (changes requested):
This PR only handles stacks from registry. what about parent reference from a URL? or from a  local path (absolute/relative URI)?

pkg/devfile/parser/parse_test.go
```diff
@@ -4317,6 +4318,52 @@ func Test_parseFromRegistry(t *testing.T) {
 	}
 }
 
+func Test_getDevfileFromDir(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "")
```
yangcao77, line 4322: similarly, use `os.TempDir()` as path, and gives it a dir name

pkg/devfile/parser/parse_test.go
```diff
@@ -4317,6 +4318,52 @@ func Test_parseFromRegistry(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("Failed to create temp dir: %s, error: %v", tempDir, err)
+	}
+	defer os.RemoveAll(tempDir)
+
+	devfile := filepath.Join(tempDir, "devfile.yaml")
+	if err := ioutil.WriteFile(devfile, []byte(""), 0666); err != nil {
```
yangcao77, line 4329: can you use the fakefs in `testingutil` instead? should avoid creating real files or dir in unit tests

pkg/util/util.go
```diff
@@ -1155,6 +1155,44 @@ func CopyFile(srcPath string, dstPath string, info os.FileInfo) error {
 	return nil
 }
 
+// CopyAllDirFiles recursively copies a source directory to a destination directory
+func CopyAllDirFiles(srcDir string, destDir string) error {
```
yangcao77, line 1159: need a unit test for this function

pkg/devfile/parser/parse.go
```diff
@@ -431,6 +454,24 @@ func parseFromRegistry(importReference v1.ImportReference, resolveCtx *resolutio
 	return DevfileObj{}, fmt.Errorf("failed to get id: %s from registry URLs provided", id)
 }
 
+func getStackFromRegistry(id, registryURL, stackDir, destDir string) error {
+	if !strings.HasPrefix(registryURL, "http://") && !strings.HasPrefix(registryURL, "https://") {
+		return fmt.Errorf("the provided registryURL: %s is not a valid URL", registryURL)
+	}
+	registryLibrary.PullStackFromRegistry(registryURL, id, stackDir, registryLibrary.RegistryOptions{})
```
yangcao77, line 461: when version is specified, stack should be `<stack>:<version>`, if no version specified, use `<stack>`

pkg/devfile/parser/parse.go
```diff
@@ -394,14 +399,29 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
 	return populateAndParseDevfile(d, newResolveCtx, tool, true)
 }
 
+func getFilesFromDir(srcDir, destDir string) {
```
yangcao77, line 402: this function should be under utils.go?

pkg/devfile/parser/parse.go
```diff
@@ -394,14 +399,29 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
...
+	// don't want tests to copy files back to the library repo
+	excludedPaths := []string{"/library/tests/v2/libraryTest", "/library/pkg"}
```
yangcao77, line 404: what is this for? do not put hard coded staffs just for the tests

pkg/devfile/parser/parse.go
```diff
@@ -374,6 +378,13 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
 	if !absoluteURL && curDevfileCtx.GetAbsPath() != "" {
 		newUri = path.Join(path.Dir(curDevfileCtx.GetAbsPath()), uri)
 		d.Ctx = devfileCtx.NewDevfileCtx(newUri)
+		if util.ValidateFile(newUri) == nil {
+			srcDir := filepath.Dir(newUri)
```
yangcao77, line 382: can use `path.Dir()` instead. no need for the filepath new import

pkg/devfile/parser/parse.go
```diff
@@ -397,11 +408,14 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
 func parseFromRegistry(importReference v1.ImportReference, resolveCtx *resolutionContextTree, tool resolverTools) (d DevfileObj, err error) {
 	id := importReference.Id
 	registryURL := importReference.RegistryUrl
+	destDir := d.Ctx.GetAbsPath()
```
yangcao77, line 411: path.Dir(d.Ctx.GetAbsPath())

pkg/devfile/parser/parse.go
```diff
@@ -441,6 +456,15 @@ func getDevfileFromRegistry(id, registryURL, version string) ([]byte, error) {
 	return util.HTTPGetRequest(param, 0)
 }
 
+func getStackFromRegistry(id, registryURL, destDir string) {
+	stackDir, _ := ioutil.TempDir(os.TempDir(), fmt.Sprintf("stack-%s", id))
+	defer os.RemoveAll(stackDir)
+
+	registryLibrary.PullStackFromRegistry(registryURL, id, stackDir, registryLibrary.RegistryOptions{})
```
yangcao77, line 463: error handleing

pkg/devfile/parser/parse.go
```diff
@@ -441,6 +456,15 @@ func getDevfileFromRegistry(id, registryURL, version string) ([]byte, error) {
 	return util.HTTPGetRequest(param, 0)
 }
 
+func getStackFromRegistry(id, registryURL, destDir string) {
+	stackDir, _ := ioutil.TempDir(os.TempDir(), fmt.Sprintf("stack-%s", id))
```
yangcao77, line 460: error handling

pkg/devfile/parser/parse.go
```diff
@@ -441,6 +456,15 @@ func getDevfileFromRegistry(id, registryURL, version string) ([]byte, error) {
 	return util.HTTPGetRequest(param, 0)
 }
 
+func getStackFromRegistry(id, registryURL, destDir string) {
```
yangcao77, line 459: this function should return error, do not ignore those errors. tools and users need to know if library failed to get the parent.

pkg/devfile/parser/parse.go
```diff
@@ -441,6 +456,15 @@ func getDevfileFromRegistry(id, registryURL, version string) ([]byte, error) {
...
+	stackDir, _ := ioutil.TempDir(os.TempDir(), fmt.Sprintf("stack-%s", id))
+	defer os.RemoveAll(stackDir)
+
+	registryLibrary.PullStackFromRegistry(registryURL, id, stackDir, registryLibrary.RegistryOptions{})
+	util.CopyAllDirFiles(stackDir, destDir)
```
yangcao77, line 464: error handling

pkg/devfile/parser/parse.go
```diff
@@ -374,6 +378,13 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
 	if !absoluteURL && curDevfileCtx.GetAbsPath() != "" {
 		newUri = path.Join(path.Dir(curDevfileCtx.GetAbsPath()), uri)
 		d.Ctx = devfileCtx.NewDevfileCtx(newUri)
+		if util.ValidateFile(newUri) == nil {
```
yangcao77, line 381: error handling. user should know if the parent folder (not exist/not readable/not writable ext. )

pkg/devfile/parser/parse.go
```diff
@@ -374,6 +378,13 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
...
+			srcDir := filepath.Dir(newUri)
+			destDir := filepath.Dir(curDevfileCtx.GetAbsPath())
+			if srcDir != destDir {
+				util.CopyAllDirFiles(srcDir, destDir)
```
yangcao77, line 385: error handling.

pkg/devfile/parser/parse.go
```diff
@@ -392,10 +397,38 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
 	if !absoluteURL && curDevfileCtx.GetAbsPath() != "" {
 		newUri = path.Join(path.Dir(curDevfileCtx.GetAbsPath()), uri)
 		d.Ctx = devfileCtx.NewDevfileCtx(newUri)
+		if !strings.HasSuffix(newUri, ".yaml") {
```
yangcao77, line 400: I'm wondering if it's reasonable to check if it's a `.yaml` format file. we've never discussed this in detail before. what if the devfile path provided contains yaml content, but it does not end with `.yaml`?

pkg/devfile/parser/parse.go
```diff
@@ -392,10 +397,38 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
...
 	} else if absoluteURL {
 		// absolute URL address
 		newUri = uri
 		d.Ctx = devfileCtx.NewURLDevfileCtx(newUri)
+		if strings.Contains(newUri, "raw.githubusercontent.com") {
+			urlComponents, err := util.GetGitUrlComponentsFromRaw(newUri)
+			if err != nil {
+				return DevfileObj{}, err
+			}
+			if !strings.Contains(urlComponents["file"], "devfile.yaml") {
```
yangcao77, line 423: you don't need to check the devfile name here. can be any name, as long as it contains a valid devfile content. and the parser will error out later if it is not a valid devfile

pkg/devfile/parser/parse.go
```diff
@@ -392,10 +397,38 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
...
+				return DevfileObj{}, fmt.Errorf("%s is not a devfile", newUri)
+			}
+			destDir := path.Dir(curDevfileCtx.GetAbsPath())
+			err = getStackFromGit(urlComponents, destDir)
+			if err != nil {
+				return DevfileObj{}, err
+			}
+		}
 	} else if curDevfileCtx.GetURL() != "" {
```
yangcao77, line 432: what about this case? consider a nested parent reference, for example
main devfile has a parent reference `somegithuburl.devfile.yaml`
and the 1st levell parent devfile from `somegithuburl.devfile.yaml` also has a second level parent reference `subfolder/devfile.yaml`.

pkg/util/util.go
```diff
@@ -1144,6 +1150,49 @@ func ValidateFile(filePath string) error {
 	return nil
 }
 
+// GetGitUrlComponentsFromRaw converts a raw GitHub file link to a map of the url components
+func GetGitUrlComponentsFromRaw(rawGitURL string) (map[string]string, error) {
+	err := ValidateURL(rawGitURL)
+	if err != nil {
+		return nil, err
+	}
+	if !strings.Contains(rawGitURL, "raw.githubusercontent.com") {
```
yangcao77, line 1159: you have checked this before calling the function, no need to check it again

pkg/util/util.go
```diff
@@ -1144,6 +1150,49 @@ func ValidateFile(filePath string) error {
...
+		return nil, errors.Errorf("%s is not a raw github url", rawGitURL)
+	}
+
+	u, _ := url.Parse(rawGitURL)
+	urlPath := strings.SplitN(u.Host+u.Path, "/", 5)
+
+	urlComponents := map[string]string{
```
yangcao77, line 1166: may want to add a short comment with an example raw giturl for better understanding

yangcao77 (commented):
add more comments, so other prople will have better understanding of the code for future maintenance.

pkg/devfile/parser/parse.go
kim-tsao: I remember an issue got opened several months ago saying we should support `.yml` extensions too https://github.com/devfile/api/issues/744

mike-hoang: I removed this check as it’s not needed here

I think the file format is being checked in the context here: https://github.com/devfile/library/blob/main/pkg/devfile/parser/context/context.go#L97-L98

pkg/devfile/parser/parse.go
mike-hoang: removed

pkg/devfile/parser/parse.go
mike-hoang: It looks like this scenario is already covered:

```yaml
# main devfile
parent:
    uri: https://raw.githubusercontent.com/mike-hoang/registry-support/API-46/tests/registry/stacks/nodejs-parent/devfile.yaml
```

```yaml
# nodejs-parent/devfile.yaml
# contains file: parent/testing.txt
parent:
    uri: ./parent/devfile.yaml
```

dest dir contains testing.txt after parsing

pkg/util/util.go
mike-hoang: removed

pkg/util/util.go
mike-hoang: added comment

pkg/devfile/parser/parse.go
yangcao77: this case works is because your 2nd parent is under the same level (or cases under lower level sub-dirs also will work), and when parsing the first parent, it is already downloaded as first parent's resource. 

if the 2nd parent is from a upper level dir, the resources are not going to be downloaded. i.e.
```
first parent: somegiturl.com/nodejs-parent/devfile.yaml
second parent: somegiturl.com/nested-parent/devfile.yaml


parent:
    uri: ../nested-parent/devfile.yaml
```

pkg/devfile/parser/parse_test.go
```diff
@@ -3080,8 +3073,8 @@ func Test_parseParentAndPlugin_RecursivelyReference(t *testing.T) {
 		}
 
 		err := parseParentAndPlugin(devFileObj, &resolutionContextTree{}, tool)
-		// devfile has a cycle in references: main devfile -> uri: http://127.0.0.1:8080 -> name: testcrd, namespace: defaultnamespace -> id: nodejs, registryURL: http://127.0.0.1:8090 -> uri: http://127.0.0.1:8080
-		expectedErr := fmt.Sprintf("devfile has an cycle in references: main devfile -> uri: %s%s -> name: %s, namespace: %s -> id: nodejs, registryURL: %s%s -> uri: %s%s", httpPrefix, uri1, name, namespace,
+		// devfile has an cycle in references: main devfile -> uri: http://127.0.0.1:8080 -> name: testcrd, namespace: defaultnamespace -> uri: http://127.0.0.1:8090 -> uri: http://127.0.0.1:8080
```
yangcao77, line 3076:
```suggestion
		// devfile has a cycle in references: main devfile -> uri: http://127.0.0.1:8080 -> name: testcrd, namespace: defaultnamespace -> uri: http://127.0.0.1:8090 -> uri: http://127.0.0.1:8080
```

pkg/devfile/parser/parse_test.go
```diff
@@ -4263,18 +4007,7 @@ func Test_parseFromRegistry(t *testing.T) {
...
-			importReference: v1.ImportReference{
-				ImportReferenceUnion: v1.ImportReferenceUnion{
-					Id: registryId,
-				},
-				Version:     "1.1.0",
-				RegistryUrl: httpPrefix + registry,
+				RegistryUrl: stagingRegistry,
 			},
 		},
 		{
```
yangcao77, line 4013: you may also want to add a failure case that failed to download resources

pkg/devfile/parser/parse_test.go
```diff
@@ -4320,9 +4053,17 @@ func Test_parseFromRegistry(t *testing.T) {
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			gotErr := false
 			got, err := parseFromRegistry(tt.importReference, &resolutionContextTree{}, tt.tool)
+			if err != nil {
+				gotErr = true
+			}
 			if (err != nil) != (tt.wantErr != nil) {
 				t.Errorf("Test_parseFromRegistry() unexpected error: %v, wantErr %v", err, tt.wantErr)
+			} else if strings.Contains(tt.name, "should be able to parse") {
```
yangcao77, line 4063: what is this for?  the test should always return a failure when the `err!=nil` and `tt.wantErr!=nil` does not match. which has been checked already

pkg/devfile/parser/parse.go
```diff
@@ -412,9 +437,32 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
 	return populateAndParseDevfile(d, newResolveCtx, tool, true)
 }
 
+func getStackFromGit(gitUrlComponents map[string]string, destDir string) error {
```
yangcao77, line 440: can you add unit tests for this new function being introduced?

pkg/devfile/parser/parse_test.go
```diff
@@ -4263,18 +4007,7 @@ func Test_parseFromRegistry(t *testing.T) {
 					Id: registryId,
 				},
 				Version:     "latest",
-				RegistryUrl: httpPrefix + registry,
-			},
-		},
-		{
-			name:        "should be able to parse from provided registryUrl with version specified",
```
yangcao77, line 4270: this test should not be deleted

pkg/util/util_test.go
```diff
@@ -1042,6 +1149,52 @@ func TestCopyFile(t *testing.T) {
...
+			destDir: destDir,
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotErr := false
+			err := copyAllDirFilesOnFS(tt.srcDir, tt.destDir, fs)
+			if err != nil {
```
yangcao77, line 1187:
```suggestion
			if (err != nil) != (tt.wantErr != nil)
```

pkg/util/util_test.go
```diff
@@ -981,7 +1017,78 @@ func TestValidateFile(t *testing.T) {
...
+			destDir:          filepath.Join(os.TempDir(), "nonexistent"),
+			wantErr:          true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotErr := false
+			err := CloneGitRepo(tt.gitUrlComponents, tt.destDir)
+			if err != nil {
```
yangcao77, line 1081: if (err != nil) != (tt.wantErr != nil)

pkg/util/util_test.go
```diff
@@ -981,7 +1017,78 @@ func TestValidateFile(t *testing.T) {
...
+			url:     invalidUrl,
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotErr := false
+			_, err := GetGitUrlComponentsFromRaw(tt.url)
+			if err != nil {
```
yangcao77, line 1045:
```suggestion
			if (err != nil) != (tt.wantErr != nil)
```

pkg/util/util_test.go
```diff
@@ -981,7 +1017,78 @@ func TestValidateFile(t *testing.T) {
...
+		"branch":   "nonexistent",
+	}
+
+	tests := []struct {
+		name             string
+		gitUrlComponents map[string]string
+		destDir          string
+		wantErr          bool
+	}{
+		{
```
yangcao77, line 1069: add a happypath unit test

pkg/util/util_test.go
```diff
@@ -1042,6 +1149,52 @@ func TestCopyFile(t *testing.T) {
 	}
 }
 
+func TestCopyAllDirFiles(t *testing.T) {
+	fs := filesystem.NewFakeFs()
+	fs.MkdirAll("/path/to/src", 0755)
+	fs.MkdirAll("/path/to/dest", 0755)
+	fs.WriteFile("/path/to/src/devfile.yaml", []byte(""), 0755)
+	fs.WriteFile("/path/to/src/file.txt", []byte(""), 0755)
```
yangcao77, line 1157: you may want to add subfolders as well for the test.

yangcao77 (changes requested):

pkg/devfile/parser/parse.go
```diff
@@ -392,29 +395,78 @@ func parseFromURI(importReference v1.ImportReference, curDevfileCtx devfileCtx.D
...
+		} else if curDevfileCtx.GetURL() != "" {
+			// relative path to a URL
+			u, err := url.Parse(curDevfileCtx.GetURL())
+			if err != nil {
+				return DevfileObj{}, err
+			}
+			u.Path = path.Join(u.Path, uri)
+			newUri = u.String()
+		} else {
+			return DevfileObj{}, fmt.Errorf("devfile context is missing absolute url and relative path to url")
```
yangcao77, line 422:
```suggestion
			return DevfileObj{}, fmt.Errorf("failed to resolve parent uri, devfile context is missing absolute url and relative path to url. %s", resolveImportReference(parent.ImportReference))
```

yangcao77 (approved):
look good to me. thanks for the change

PR #160, opened 2022-12-20
adding support for public and private git providers

mike-hoang:
Signed-off-by: Michael Hoang <mhoang@redhat.com>

### What does this PR do?:
Adds a parser to handle Github, GitLab, and Bitbucket urls along with updating `CloneGitRepo` to support private repos with personal auth tokens.

fixes https://github.com/devfile/api/issues/984

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

- [ ] Gosec scans


pkg/util/git.go
```diff
@@ -0,0 +1,276 @@
...
+				case "blob":
+					g.IsFile = true
+				}
+				g.Branch = splitUrl[3]
+				g.Path = splitUrl[4]
+			}
+		}
+	}
+
+	if !isGitUrlPublic(g, c) {
```
kim-tsao, line 134: We shouldn't rely on a false value to assume it's a private repo.  False can mean the URL is unreachable/non-existent.  We also need to consider error scenarios where users may be providing a private URL but have given us a bad token or did not set the env var.   

- If the token does not exist, we should issue an error asking users to either check their URL or set the token if private repo was intended.  

- If the token exists, we should check if it's reachable.  If not, issue an error and stop the processing

pkg/util/git.go
```diff
@@ -0,0 +1,276 @@
...
+			ext := filepath.Ext(g.Path)
+			if ext != "" {
+				g.IsFile = true
+			}
+		} else {
+			err = fmt.Errorf("url path should contain 'blob' or 'tree' or 'raw', received: %s", url.Path[1:])
+		}
+	}
+
+	if !isGitUrlPublic(g, c) {
```
kim-tsao, line 178: See comment from above

pkg/util/git.go
```diff
@@ -0,0 +1,276 @@
...
+				}
+			} else {
+				err = fmt.Errorf("url path should contain 'raw' or 'src', received: %s", url.Path[1:])
+			}
+		} else {
+			err = fmt.Errorf("url path should contain path to directory or file, received: %s", url.Path[1:])
+		}
+	}
+
+	if !isGitUrlPublic(g, c) {
```
kim-tsao, line 218: See comment from above

pkg/util/git.go
```diff
@@ -0,0 +1,276 @@
...
+		cloneOptions.Auth = &githttp.BasicAuth{
+			// go-git auth allows username to be anything except
+			// an empty string for GitHub and GitLab, however requires
+			// for Bitbucket to be "x-token-auth"
+			Username: "x-token-auth",
+			Password: g.token,
+		}
+	}
+
+	_, err := gitpkg.PlainClone(destDir, false, cloneOptions)
```
kim-tsao, line 270: We had problems using the go-git cloning function in HAS for large repos.  See https://github.com/redhat-appstudio/application-service/pull/204.  It'll be good if we can adopt the same workaround to use the git cli instead

pkg/util/git.go
```diff
@@ -0,0 +1,276 @@
+//
+// Copyright 2022 Red Hat, Inc.
```
maysunfaisal, line 2:
```suggestion
// Copyright 2023 Red Hat, Inc.
```

pkg/util/git.go
```diff
@@ -0,0 +1,276 @@
...
+type GitUrl struct {
+	Protocol string
+	Host     string
+	Owner    string
+	Repo     string
+	Branch   string
+	Path     string
+	token    string
+	IsFile   bool
+}
```
maysunfaisal, line 50: A one line desc for each would be nice

pkg/util/git.go
```diff
@@ -0,0 +1,276 @@
...
+func CloneGitRepo(g GitUrl, destDir string) error {
+	var cloneOptions *gitpkg.CloneOptions
+
+	host := g.Host
+	if host == RawGitHubHost {
+		host = GitHubHost
+	}
+
+	repoUrl := fmt.Sprintf("%s://%s/%s/%s.git", g.Protocol, host, g.Owner, g.Repo)
+	branch := fmt.Sprintf("refs/heads/%s", g.Branch)
```
maysunfaisal, line 251: is the branch format universal for all Git repos?

pkg/util/git.go
```diff
@@ -0,0 +1,276 @@
...
+
+	cloneOptions = &gitpkg.CloneOptions{
+		URL:           repoUrl,
+		ReferenceName: plumbing.ReferenceName(branch),
+		SingleBranch:  true,
+		Depth:         1,
```
maysunfaisal, line 257: So this fetches a commit of up to 1, wonder what's the difference in cloning when you have it at 1 vs something more than 1...

pkg/util/git.go
```diff
@@ -0,0 +1,276 @@
...
+
+const (
+	GitHubHost    string = "github.com"
+	RawGitHubHost string = "raw.githubusercontent.com"
+	GitLabHost    string = "gitlab.com"
+	BitbucketHost string = "bitbucket.org"
+
+	GitHubToken    string = "GITHUB_TOKEN"
+	GitLabToken    string = "GITLAB_TOKEN"
+	BitbucketToken string = "BITBUCKET_TOKEN"
```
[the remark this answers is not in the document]
mike-hoang: Not sure if there's a better place for it, but added to readme for now

pkg/util/git.go
mike-hoang: Copied the workaround and it looks good

pkg/util/git.go
mike-hoang: Yeah the format works across all three providers, but I ended up not using the go-git package anymore

pkg/util/git.go
mike-hoang: I think increasing the depth would increase the size of the clone (and the git log)

pkg/util/git.go
mike-hoang: Not sure if parsing the url should make http calls, so I moved the token check to the cloning function

README.md
```diff
@@ -167,6 +167,21 @@ The function documentation can be accessed via [pkg.go.dev](https://pkg.go.dev/g
...
+
+   // Clone the repo to a destination dir
+   err = util.CloneGitRepo(gitUrl, destDir)
+   ```
+   If repository is private, set the correct environment variables
+   ```shell
+   # credentials for private repositories
+   export GITHUB_TOKEN=<account_token>
+   export GITLAB_TOKEN=<account_token>
+   export BITBUCKET_TOKEN=<account_token>
```
kim-tsao, line 183: We also have another scenario where private repos are supported - if a devfile has a parent uri to a private repo.  Perhaps call out private repo support at the top of the README and say tokens need to be set if 1) you're a tooling client calling the library APIs and 2) if you need to parse your devfile containing a parent devfile from a private repo

README.md
```diff
@@ -167,6 +167,21 @@ The function documentation can be accessed via [pkg.go.dev](https://pkg.go.dev/g
    resources, err := ParseKubernetesYaml(values)
    ```
 
+8. To get resources from Git repositories
+   ```go
+   // Parse the repo url
+   gitUrl, err := util.ParseGitUrl(url)
```
kim-tsao, line 173: Specify the expected format of the  `url` arg.  Since it can vary depending on what repo type, we should be clear as to what formats we support

kim-tsao: Additionally, ensure this information gets added to the API description as well.

pkg/util/git.go
```diff
@@ -215,61 +198,51 @@ func parseBitbucketUrl(g GitUrl, url *url.URL, c http.Client) (GitUrl, error) {
...
-	if res, err := c.Get(repo); err != nil {
-		return false
-	} else if res.StatusCode == http.StatusOK {
-		return true
-	}
-	return false
-}
-
-// CloneGitRepo clones a GitHub Repo to a destination directory
+// CloneGitRepo clones a git repo to a destination directory
```
kim-tsao, line 204: Describe the expected format for `destDir` and ensure it gets validated.   If invalid, issue a user error

pkg/util/git.go
```diff
@@ -215,61 +198,51 @@ func parseBitbucketUrl(g GitUrl, url *url.URL, c http.Client) (GitUrl, error) {
...
-			// go-git auth allows username to be anything except
-			// an empty string for GitHub and GitLab, however requires
-			// for Bitbucket to be "x-token-auth"
-			Username: "x-token-auth",
-			Password: g.token,
+	// check if the git repo is public
+	_, err := HTTPGetRequest(params, 0)
+	if err != nil {
+		// private git repo requires authentication
+		isPublic = false
```
kim-tsao, line 222: We can't rely on an error to say the the repo is private.  We need to make a GET call with auth to determine that

pkg/util/git.go
```diff
@@ -215,61 +198,51 @@ func parseBitbucketUrl(g GitUrl, url *url.URL, c http.Client) (GitUrl, error) {
...
+		repoUrl = fmt.Sprintf("%s://token:%s@%s/%s/%s.git", g.Protocol, g.token, host, g.Owner, g.Repo)
```
kim-tsao, line 223: `g.token` may be empty at this point if user did set the env var.  Add a check to make sure it's not empty before the lines 223 and 225

pkg/util/git.go
```diff
@@ -215,61 +198,51 @@ func parseBitbucketUrl(g GitUrl, url *url.URL, c http.Client) (GitUrl, error) {
...
+	_, err = c.CombinedOutput()
 	if err != nil {
-		return err
+		if !isPublic {
+			if g.token == "" {
+				return fmt.Errorf("failed to clone repo without a token, ensure that a token is set if the repo is private. error: %v", err)
+			}
+			return fmt.Errorf("failed to clone repo with token, ensure that the url and token is correct. error: %v", err)
+		}
+		return fmt.Errorf("failed to clone repo, ensure that the url is correct. error: %v", err)
```
kim-tsao, line 245: error out before cloning by testing repo with and without auth (lines 219-225) - we don't want to make bad clone calls.

kim-tsao (commented):
The implementation looks incomplete for main devfile support.  If you walk through the code starting from `ParseDevfile` > `populateAndParseDevfile` > `PopulateFromURL`> `SetDevfileContent`, there is no token processing in those methods.   You will need to modify the `DevfileCtx` structure to include a token field which will need to be set before parsing begins.    I see you updated the `DownloadInMemory` function to process a token but if it's not being passed in, it's effectively the same as an empty string (no token) so private repo access will fail.  

Can you also ensure the cloned repo gets cleaned up once we're done with it?  We don't want to litter the disk/volumes with temp files.

mike-hoang:
> The implementation looks incomplete for main devfile support. If you walk through the code starting from `ParseDevfile` > `populateAndParseDevfile` > `PopulateFromURL`> `SetDevfileContent`, there is no token processing in those methods. You will need to modify the `DevfileCtx` structure to include a token field which will need to be set before parsing begins. I see you updated the `DownloadInMemory` function to process a token but if it's not being passed in, it's effectively the same as an empty string (no token) so private repo access will fail.

Thanks for the catch! Added the token processing for main devfiles

> Can you also ensure the cloned repo gets cleaned up once we're done with it? We don't want to litter the disk/volumes with temp files.

Cloned repos are currently removed after use

pkg/util/util.go
```diff
@@ -1083,17 +1084,46 @@ func DownloadFileInMemory(url string) ([]byte, error) {
...
 		return nil, err
 	}
 
+	if IsGitProviderRepo(url) {
+		url = g.GitRawFileAPI()
+		req, err = http.NewRequest("GET", url, nil)
+		if err != nil {
+			return nil, err
+		}
+		if !g.IsPublic(params.Timeout) {
```
kim-tsao, line 1118: similar comment to above, we don't need to check

pkg/util/git.go
```diff
@@ -0,0 +1,324 @@
...
+			if len(splitUrl) == 5 {
+				switch splitUrl[2] {
+				case "tree":
+					g.IsFile = false
+				case "blob":
+					g.IsFile = true
+				}
+				g.Branch = splitUrl[3]
+				g.Path = splitUrl[4]
+			}
```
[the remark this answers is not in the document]
mike-hoang: If we error out, we won't be able to get repo url that don't have point to a file/directory (i.e. https://github.com/devfile/library)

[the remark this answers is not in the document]
kim-tsao: Are repo urls that do not have the matching length valid?  If not, then we should error out.  You handled the error in the `parseGitHubUrl` method when length != 4 so it looked inconsistent that you weren't doing it here

pkg/devfile/parser/parse_test.go
```diff
@@ -4230,7 +4227,7 @@ func Test_parseFromURI_PrivateToken(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := parseFromURI(tt.importReference, tt.curDevfileCtx, &resolutionContextTree{}, resolverTools{token: tt.token})
+			got, err := parseFromURI(tt.importReference, tt.curDevfileCtx, &resolutionContextTree{}, resolverTools{})
```
kim-tsao, line 4230: The purpose of this test is not to validate if the token got set since this is already known and it is part of your setup.  What should be tested are your 3 supported scenarios:

1.  Verify that you can parse the devfile contents for a private main devfile URL
2.  Verify that you can parse the devfile contents for your private main devfile with a private parent reference
3.  Verify you can can parse the devfile contents for your private main devfile with public parent reference 

Those are the happy path scenarios.  There should also be negative path testing which can include: 

1. Parsing a private main devfile without a valid token
2. Parsing a bad devfile URL (malformed but vendors are valid) with/without a valid token - does this fail gracefully?
3.  Parsing a bad devfile URL (unsupported vendors), etc

These are just some examples that you can start with.  

This means you will need to set up your test servers to handle requests with tokens.  You can take a look at this [mock server](https://github.com/redhat-appstudio/application-service/blob/main/pkg/github/mock.go#L46) for some ideas.

pkg/util/git.go
```diff
@@ -117,15 +119,26 @@ func (g *GitUrl) parseGitHubUrl(url *url.URL) error {
...
-				switch splitUrl[2] {
-				case "tree":
-					g.IsFile = false
-				case "blob":
-					g.IsFile = true
-				}
 				g.Branch = splitUrl[3]
 				g.Path = splitUrl[4]
+			} else {
+				err = fmt.Errorf("url path should contain <owner>/<repo>/<tree or blob>/<branch>/<path/to/file/or/directory>, recieved: %s", url.Path[1:])
```
kim-tsao, line 141: typo on `recieved`

pkg/devfile/parser/context/context.go
```diff
@@ -45,13 +46,19 @@ type DevfileCtx struct {
...
 
-	//url path of the devfile
+	// url path of the devfile
 	url string
 
+	// token is a personal access token used with a private git repo URL
+	token string
+
+	// Git is an interface used for git urls
+	Git git.IGitUrl
```
yangcao77, line 56: Do not expose the value. similar as other params. create `Set..()` and `Get...()` methods if you want to access from external.

pkg/git/git.go
```diff
@@ -0,0 +1,428 @@
...
+	} else if parsedUrl.Host == BitbucketHost {
+		err = g.parseBitbucketUrl(parsedUrl)
+	} else {
+		err = fmt.Errorf("url host should be a valid GitHub, GitLab, or Bitbucket host; received: %s", parsedUrl.Host)
+	}
+
+	return g, err
+}
+
+func (g *Url) GetProtocol() string {
```
yangcao77, line 94: Do we need those helper functions? aren't those already exposed?

pkg/git/git.go
```diff
@@ -0,0 +1,428 @@
...
+	IsPublic(httpTimeout *int) bool
+	DownloadGitRepoResources(url string, destDir string, httpTimeout *int, token string) error
+}
+
+type Url struct {
+	Protocol string // URL scheme
+	Host     string // URL domain name
+	Owner    string // name of the repo owner
+	Repo     string // name of the repo
+	Branch   string // branch name
```
yangcao77, line 48: only branch is supported? can we do revision instead? I think tag/commitID also need to be supported.

pkg/git/git.go
```diff
@@ -0,0 +1,428 @@
...
+	var splitUrl []string
+	var err error
+
+	g.Protocol = url.Scheme
+	g.Host = url.Host
+
+	if g.Host == RawGitHubHost {
+		g.IsFile = true
+		// raw GitHub urls don't contain "blob" or "tree"
+		splitUrl = strings.SplitN(url.Path[1:], "/", 4)
```
yangcao77, line 225: better adds an example here for better code understanding

pkg/git/git.go
```diff
@@ -0,0 +1,428 @@
...
+			g.Path = splitUrl[3]
+		} else {
+			// raw GitHub urls have to be a file
+			err = fmt.Errorf("raw url path should contain <owner>/<repo>/<branch>/<path/to/file>, received: %s", url.Path[1:])
+		}
+		return err
+	}
+
+	if g.Host == GitHubHost {
+		splitUrl = strings.SplitN(url.Path[1:], "/", 5)
```
yangcao77, line 239: better adds an example here in the comment for better code understanding

pkg/git/git.go
mike-hoang: Yeah I wasn't sure what the best way to name this was so I just chose `Branch`, but it does support them all:

https://github.com/devfile/library/tree/0ce592a416fb185564516353891a45016ac7f671
https://github.com/devfile/library/blob/0ce592a416fb185564516353891a45016ac7f671/devfile.yaml

- 	`branch = 0ce592a416fb185564516353891a45016ac7f671`

https://github.com/devfile/library/tree/main
https://github.com/devfile/library/blob/main/devfile.yaml

- 	`branch = main`

https://github.com/devfile/library/tree/v2.2.0
https://github.com/devfile/library/blob/v2.2.0/devfile.yaml

- 	`branch = v2.2.0`

kim-tsao (commented):
Can you update the license headers on all modified files to the current year?

pkg/git/git.go
```diff
@@ -0,0 +1,367 @@
...
+)
+
+const (
+	GitHubHost    string = "github.com"
+	RawGitHubHost string = "raw.githubusercontent.com"
+	GitLabHost    string = "gitlab.com"
+	BitbucketHost string = "bitbucket.org"
+)
+
+type Url struct {
```
kim-tsao, line 33: Do we want to rename this back to GitUrl to be consistent with the comments,  mock type and util functions?  Also, the name `Url` is pretty generic

pkg/devfile/parser/parse_test.go
```diff
@@ -4155,6 +4162,280 @@ func Test_parseFromURI(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			downloadGitRepoResources = mockDownloadGitRepoResources(tt.gitUrl)
+			got, err := parseFromURI(tt.importReference, *tt.curDevfileCtx, &resolutionContextTree{}, resolverTools{})
+			if (err != nil) != (tt.wantError != nil) {
+				t.Errorf("Unexpected error: %v, wantErr %v", err, tt.wantError)
+			} else if err == nil && !reflect.DeepEqual(got.Data, tt.wantDevFile.Data) {
```
kim-tsao, line 4385: It's not just the parsed devfile we want to validate, we also want to validate your new changes too, specifically the output of the clone function.  

In the mockExecute function, you return content "test" if a repo is successfully cloned.  If it has a revision that you can switch to, the expected content then becomes "git switched to revision". You should test out these combinations: 
 1.  If the git URL is valid and has no revisions, then validate the output is "test"
 2. If the git URL has an invalid revision then make sure destination directory is empty since the entire operation should fail and you will need to clean up.  I think there should be a parse error here too if the parent devfile can't be determined.  Just to point out for mocking purposes,  we are forcing `tt.gitUrl` to be the parent URL since the real URL we set up as the import reference is not a valid git provider (it's just a localhost server)
 3. If the git URL has a valid revision, then the content should be "git switched to revision"

pkg/git/git.go
```diff
@@ -0,0 +1,367 @@
...
+		} else {
+			return fmt.Errorf("failed to clone repo with token, ensure that the url and token is correct. error: %v", err)
+		}
+	}
+
+	if g.Revision != "" {
+		_, err := execute(destDir, "git", "switch", "--detach", "origin/"+g.Revision)
+		if err != nil {
+			return fmt.Errorf("failed to switch repo to revision. repo dir: %v, revision: %v", destDir, g.Revision)
+		}
```
kim-tsao, line 144: make sure local files are cleaned up if revision switch fails especially if previous clone command was successful

README.md
```diff
@@ -13,6 +13,32 @@ The Devfile Parser library is a Golang module that:
...
+Note: The url must also be set with a supported git provider repo url.
+
+For more information about personal access tokens:
+1. [GitHub docs](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)
+2. [GitLab docs](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html#create-a-personal-access-token)
+3. [Bitbucket docs](https://support.atlassian.com/bitbucket-cloud/docs/repository-access-tokens/)
+
+[1] Currently, this works under the assumption that the token can authenticate the devfile and the parent devfile; both devfiles are in the same repository.
+[2] In this scenario, the token will be used to authenticate the main devfile.
+
```
kim-tsao, line 41: Can you also update the README to explain the changes that were made in the previous PR (and now expanded by this PR)

https://github.com/devfile/library/pull/139

pkg/devfile/parser/parse.go
```diff
@@ -46,6 +48,61 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	DevfileName = "devfile.yaml"
```
kim-tsao, line 52: You can make this package scoped if it's not used anywhere else i.e. `devfileName` instead of `DevfileName`

kim-tsao: Actually, you can delete this and use the existing `OutputDevfileYamlPath` constant instead

pkg/git/mock.go
```diff
@@ -0,0 +1,137 @@
...
+	if cmd == GitCommand {
+		if len(args) > 0 && args[0] == "clone" {
+			u, _ := url.Parse(args[1])
+			password, hasPassword := u.User.Password()
+
+			// private repository
+			if hasPassword {
+				switch password {
+				case "valid-token":
+					_, err := os.Create(filepath.Clean(baseDir) + "/private-repo-resource")
```
kim-tsao, line 51: You can use our existing utility functions for this: https://github.com/devfile/library/blob/main/pkg/testingutil/filesystem/filesystem.go

pkg/git/mock.go
```diff
@@ -0,0 +1,137 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to create test resource: %v", err)
+			}
+			return []byte(""), nil
+		}
+
+		if len(args) > 0 && args[0] == "switch" {
+			revision := strings.TrimPrefix(args[2], "origin/")
+			if revision != "invalid-revision" {
+				_, err := os.Create(filepath.Clean(baseDir) + "/valid-git-switch")
```
kim-tsao, line 71: When we do a revision switch, you can have the same resource but different file contents which is why I suggested to write out the byte contents into a file and then check if the file exists with the right content.

pkg/git/git_test.go
```diff
@@ -0,0 +1,448 @@
...
+				Repo:     "library",
+				Revision: "main",
+				Path:     "devfile.yaml",
+				IsFile:   true,
+			},
+		},
+		{
+			name:    "should fail with missing GitHub repo",
+			url:     "https://github.com/devfile",
+			wantErr: "url path should contain <user>/<repo>*",
```
kim-tsao, line 118: small nit, I would make these messages string constants

PR #165, opened 2023-02-20
Feature: pod security admission

feloy:
### What does this PR do?:

This PR adds the possibility to get a patched PodTemplate based on a Pod Security Admission policy, so the Pod can be accepted by the Pod Security Admission controller.

Prior to being able to do this, some refactoring has been done:

- updating the k8s dependencies to latest version, to be able to use the latest version of k8s.io/pod-security-admission
- update to go 1.19, to be able to use the latest k8s dependencies
- A new method `GetPodTemplateSpec` has been added, that will handle the Pod Security patching and the container/pod overrides in an atomic way (instead of making them in different places), to be used instead of GetContainers/GetInitContainers
- GetDeployment now accepts a podSecurityTemplate instead of containers, initContainers and volumes (these 3 values are now marked as deprecated)

https://github.com/redhat-developer/odo/issues/6339

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

- [ ] Gosec scans


pkg/devfile/generator/generators.go
```diff
@@ -168,25 +169,28 @@ func GetInitContainers(devfileObj parser.DevfileObj) ([]corev1.Container, error)
...
+	}
 
 	podTemplateSpecParams := podTemplateSpecParams{
-		ObjectMeta:     deployParams.ObjectMeta,
-		InitContainers: deployParams.InitContainers,
-		Containers:     deployParams.Containers,
-		Volumes:        deployParams.Volumes,
+		ObjectMeta:     podTemplateParams.ObjectMeta,
+		InitContainers: initContainers,
+		Containers:     containers,
```
rm3l, line 193: I think we should also set the `Volumes` field here, no? Perhaps by exposing it in the `PodTemplateParams` struct..
Maybe I missed it, but I couldn't see how Volumes could be defined easily.

feloy: On Odo, they are set afterwards, on the returned podtemplatespec. That seems easier, as we also want to set the volumemounts in the containers at the same time.

elsony:
To maintain backward compatibility, we need to keep the old functions and create new ones to provide the new function

pkg/devfile/generator/policy.go
```diff
@@ -0,0 +1,131 @@
...
+	})
+	return podTemplateSpec
+}
+
+func patchRunAsNonRoot(podTemplateSpec *corev1.PodTemplateSpec) *corev1.PodTemplateSpec {
+	if podTemplateSpec.Spec.SecurityContext == nil {
+		podTemplateSpec.Spec.SecurityContext = &corev1.PodSecurityContext{}
+	}
+	podTemplateSpec.Spec.SecurityContext.RunAsNonRoot = pointer.Bool(true)
+	// No need to set the value as true for containers, as setting at the Pod level is sufficient
```
rm3l, line 107: To me (and per https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container), security settings at the Container level override settings made at the Pod level when there is an overlap. So I think, we don't need to set the value as `true` to containers only if they are not specified at the container level.
So we should still visit all containers and set the field to `true`, no?

rm3l (changes requested):

pkg/devfile/generator/policy.go
feloy: > To me (and per https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container), security settings at the Container level override settings made at the Pod level when there is an overlap. So I think, we don't need to set the value as `true` to containers only if they are not specified at the container level. 

I agree up to here.

> So we should still visit all containers and set the field to `true`, no?

By design, we know the field at container level will never be set (because they could be set only from the library, and they are not). For this reason, we only need to set it at pod level.

pkg/testingutil/k8sClient.go
```diff
@@ -30,7 +30,7 @@ type FakeK8sClient struct {
 	Errors                map[string]string
 }
 
-func (client *FakeK8sClient) Get(_ context.Context, namespacedName client.ObjectKey, obj client.Object) error {
+func (client *FakeK8sClient) Get(_ context.Context, namespacedName client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
```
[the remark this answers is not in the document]
feloy: This parameter has been added for Get to implement again the `Reader` interface from `controller-runtime`, which has changed. Renaming it to `_` instead.

pkg/devfile/generator/generators.go
```diff
@@ -170,23 +176,98 @@ func GetInitContainers(devfileObj parser.DevfileObj) ([]corev1.Container, error)
...
+			Replicas:          deployParams.Replicas,
+		}
+	} else {
+		if len(deployParams.InitContainers) > 0 ||
+			len(deployParams.Containers) > 0 ||
+			len(deployParams.Volumes) > 0 {
+			return nil, errors.New("InitContainers, Containers and Volumes cannot be set when PodTemplateSpec is set in parameters")
+		}
+
+		deploySpecParams = deploymentSpecParams{
```
yangcao77, line 221: dup with https://github.com/devfile/library/blob/5985a2891bffdf9e0a4dddb46a9819381ee8ff8e/pkg/devfile/generator/generators.go#L209-L212, can be move out side of the if-else block?

pkg/devfile/generator/generators.go
```diff
@@ -199,29 +280,74 @@ func GetDeployment(devfileObj parser.DevfileObj, deployParams DeploymentParams)
...
+		ComponentOptions: common.ComponentOptions{
+			ComponentType: v1.ContainerComponentType,
+		},
+	})
+	if err != nil {
+		return nil, err
 	}
 
-	return deployment, nil
+	getContainerByName := func(name string) (*corev1.Container, bool) {
```
yangcao77, line 325: you do not need to introduce a new function for this purpose. 
should use https://github.com/devfile/library/blob/5985a2891bffdf9e0a4dddb46a9819381ee8ff8e/pkg/devfile/parser/data/v2/components.go#L28, and set `DevfileOptions.FilterByName = <the containerName>`

pkg/devfile/generator/generators.go
```diff
@@ -170,23 +176,98 @@ func GetInitContainers(devfileObj parser.DevfileObj) ([]corev1.Container, error)
...
+}
+
+// GetPodTemplateSpec returns a pod template
+// The function:
+// - iterates through all container components, filters out init containers and gets corresponding containers
+// - gets the init container for every preStart devfile event
+// - patches the pod template and containers to satisfy PodSecurityAdmissionPolicy
+// - patches the pod template and containers to apply pod and container overrides
+func GetPodTemplateSpec(devfileObj parser.DevfileObj, podTemplateParams PodTemplateParams) (*corev1.PodTemplateSpec, error) {
+	containers, err := GetContainers(devfileObj, common.DevfileOptions{})
```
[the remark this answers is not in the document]
feloy: IMO, the GetContainers/GetInitContainers could be removed later from the library interface, by renaming them as getContainers/getInitContainers

feloy:
Thanks @yangcao77 for you review

yangcao77 (approved):
thanks for making this change, generally looks good to me

PR #171, opened 2023-04-18
fix the invalid kube yaml causing panic issue

yangcao77:
### What does this PR do?:

This PR updates  the kubeyaml parser to use unmarshal to ensure the kube yaml is in the expected format. to avoid the panic when trying to convert an `interface{}` into a 'map[string]interface {}'

```
panic: interface conversion: interface {} is string, not map[string]interface {} [recovered]
	panic: interface conversion: interface {} is string, not map[string]interface {}
```

for has bug: https://issues.redhat.com/browse/RHTAPBUGS-113

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

- [ ] Gosec scans

### DIFF
--- a/pkg/devfile/parser/reader.go
+++ b/pkg/devfile/parser/reader.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2022 Red Hat, Inc.
+// Copyright 2022-2023 Red Hat, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -121,7 +121,12 @@ func ParseKubernetesYaml(values []interface{}) (KubernetesResources, error) {
 			return KubernetesResources{}, err
 		}
 
-		kubernetesMap := value.(map[string]interface{})
+		// kubernetesMap := value.(map[string]interface{})
+		var kubernetesMap map[string]interface{}
+		err = k8yaml.Unmarshal(byteData, &kubernetesMap)
+		if err != nil {
+			return KubernetesResources{}, err
+		}
 		kind := kubernetesMap["kind"]
 
 		switch kind {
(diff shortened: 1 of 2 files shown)


pkg/devfile/parser/reader.go
```diff
@@ -121,7 +121,12 @@ func ParseKubernetesYaml(values []interface{}) (KubernetesResources, error) {
 			return KubernetesResources{}, err
 		}
 
-		kubernetesMap := value.(map[string]interface{})
+		// kubernetesMap := value.(map[string]interface{})
```
maysunfaisal, line 124: comment

pkg/devfile/parser/reader_test.go
```diff
@@ -72,7 +72,9 @@ func TestReadAndParseKubernetesYaml(t *testing.T) {
 		name                string
 		src                 YamlSrc
 		fs                  *afero.Afero
+		testParseYamlOnly   bool
```
maysunfaisal, line 75: where are we setting this? We are using it but i dont see it being set a value

yangcao77: https://github.com/devfile/library/blob/88946b65ac8d82e7e886b825ff76d8d1bc51893a/pkg/devfile/parser/reader_test.go#L177

pkg/devfile/parser/reader.go
yangcao77: updated

maysunfaisal (approved):

maysunfaisal:
@yangcao77 tests failing

yangcao77:
@maysunfaisal fixed the test failure. can you help review again ? thanks

PR #172, opened 2023-04-24
Allow using image names as selectors

rm3l:
### What does this PR do?:
This PR introduces a new `ImageNamesAsSelector` field in the Devfile parser args. If this field is not `nil`, the parser will try to use image names from Image components as selectors in any matching Container or Kubernetes/OpenShift components.
In this case, the parser will return a Devfile object where all matching image names have been replaced with the arguments provided in the parser args.
Replacement is done only on relative image names and the logic to check whether there is a match is based on the discussion in https://github.com/devfile/api/issues/985#issuecomment-1404939472.

This is needed by `odo` (https://github.com/redhat-developer/odo/issues/6565).

Fixes https://github.com/devfile/api/issues/985
Fixes https://github.com/redhat-developer/odo/issues/6676

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact
This is backward compatible, but the behavior of considering `imageName` as an image selector will need to be documented.

  

- [ ] Gosec scans

### DIFF
--- a/pkg/devfile/parse.go
+++ b/pkg/devfile/parse.go
@@ -109,6 +109,15 @@ func ParseDevfileAndValidate(args parser.ParserArgs) (d parser.DevfileObj, varWa
 		varWarning = variables.ValidateAndReplaceGlobalVariable(d.Data.GetDevfileWorkspaceSpec())
 	}
 
+	// Use image names as selectors after variable substitution,
+	// as users might be using variables for image names.
+	if args.ImageNamesAsSelector != nil && args.ImageNamesAsSelector.Registry != "" {
+		err = replaceImageNames(&d, args.ImageNamesAsSelector.Registry, args.ImageNamesAsSelector.Tag)
+		if err != nil {
+			return d, varWarning, err
+		}
+	}
+
 	// generic validation on devfile content
 	err = validate.ValidateDevfileData(d.Data)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/devfile/parse.go
```diff
@@ -109,6 +109,15 @@ func ParseDevfileAndValidate(args parser.ParserArgs) (d parser.DevfileObj, varWa
 		varWarning = variables.ValidateAndReplaceGlobalVariable(d.Data.GetDevfileWorkspaceSpec())
 	}
 
+	// Use image names as selectors after variable substitution,
+	// as users might be using variables for image names.
+	if args.ImageNamesAsSelector != nil && args.ImageNamesAsSelector.Registry != "" {
```
yangcao77, line 114: is `registry` a mandatory field? if it is, should mark as `required ` in the struct?

rm3l: Whoops, I missed your comment.

> is registry a mandatory field?
>
Initially, I didn't want to make it required. But with your comment, I think it does not make sense to pass a non-nil `ImageNamesAsSelector` with no registry inside.

> if it is, should mark as required  in the struct?
>
[`164ec3d` (#172)](https://github.com/devfile/library/pull/172/commits/164ec3d27a4041ede27bf7690d74a31f5459711c) makes it mandatory and checks if it is set when `ImageNamesAsSelector` is non-nil.

PR #177, opened 2023-08-04
Docker compose to devfile 

Mahajanet:
### What does this PR do?:
This PR provides library support for a project with a docker-compose file in it while supporting existing containers, single service & Multi-service with no dependencies.
We are using [Kompose](https://github.com/kubernetes/kompose) and it's framework instead of using it as a CLI. 
For the [issue](https://github.com/devfile/api/issues/1122) in reference, the testing for converting docker-compose references for inline, URI, and both absolute and relative URL cases is completed. The next steps would include: 

1. Using the apply and deploy on the kubernetes components created
2. Cleaning up the temoprary folder created for the intermediate kubernetes representation
3. Unit Testing. 

This PR fixes the issue [1122](https://github.com/devfile/api/issues/1122), which is under the Devfile to Docker Compose Epic [501](https://github.com/devfile/api/issues/501).

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [ ] Unit/Functional tests

  

- [ ] Documentation 

   

- [ ] Client Impact

  

- [ ] Gosec scans


main.go
```diff
@@ -50,8 +50,10 @@ func parserTest() {
 		fmt.Println("parsing devfile from " + os.Args[1])
 
 	} else {
+		trueFlag := true
 		args = parser.ParserArgs{
-			Path: "devfile.yaml",
+			Path:                             "devfile.yaml",
+			ConvertDockerComposeToKubernetes: &trueFlag, //setting flag to true for parser
```
amisevsk, line 56: You could use the [`k8s.io/utils/pointer`](https://pkg.go.dev/k8s.io/utils/pointer) or [`k8s.io/utils/ptr`](https://pkg.go.dev/k8s.io/utils/ptr) package for this, e.g.

```suggestion
			ConvertDockerComposeToKubernetes: pointer.Bool(true), //setting flag to true for parser
```

pkg/dockercompose/convert.go
```diff
@@ -0,0 +1,313 @@
...
+	BUILD_CONFIG ConvertBuild = "build-config"
+	NONE         ConvertBuild = "none"
+)
+
+type KubernetesController string
+
+const (
+	DEPLOYMENT             KubernetesController = "deployment"
+	DAEMONSET              KubernetesController = "daemonSet"
+	REPLICATION_CONTROLLER KubernetesController = "replicationController"
```
amisevsk, line 23: In what cases is a replicationController to be used? My understanding is that a replication controller is an older way to have the functionality provided by deployments (+ replicaSets).

pkg/devfile/parser/parse.go
```diff
@@ -914,3 +931,124 @@ func getKubernetesDefinitionFromUri(uri string, d devfileCtx.DevfileCtx) ([]byte
...
+			ComponentType: v1.ComposeComponentType,
+		},
+	}
+	//parse for docker compose components
+	dockercomposeComponents, err := devObj.Data.GetComponents(getDockercomposeOptions)
+	if err != nil {
+		return err
+	}
+	//iterate through each docker compose component
+	for _, dockercomposeComp := range dockercomposeComponents {
```
amisevsk, line 975: Consider breaking this up into multiple functions (e.g. a function for handling URIs), I'm finding it hard to parse with all the conditionals.

amisevsk (commented):
Unclear from an initial reading of this PR, but is there a way to use docker compose support without relying on a filesystem? E.g. a way to pass in a devfile and get k8s components back out of it?

Something like
```
func convertDockerComposeToKubernetes(composeComponents []devfile.Component) ([]Client.Object, error)
```

michael-valdron:
PR is stale, closing it now.

PR #178, opened 2023-08-16
Add mock interface and avoid validating tokens

kim-tsao:
### What does this PR do?:

- Expose the interface for mock testing parent URI references for external clients
- Remove duplicate util functions. There are identical copies in the util and git packages and I believe this was done in order to avoid cyclic dependencies. Rather than maintain two copies, I will refactor everything into one package (util) which this may affect our clients if they use the util methods from the git package.
- Ensure git vendor validation is in place for parent URI references. Similar to the main devfile scenario, we do not error out if absolute URL is not a supported git vendor
- Fix test cases

https://github.com/devfile/api/issues/1216

### PR acceptance criteria:

> - Check each criteria if:
>  **or**

- [x] Unit/Functional tests

  

- [ ] Documentation 

   

- [x] Client Impact

  

- [x] Gosec scans


pkg/devfile/parser/parse.go
```diff
@@ -217,6 +165,11 @@ func ParseDevfile(args ParserArgs) (d DevfileObj, err error) {
 		d.Ctx.SetToken(args.Token)
 	}
 
+	args.DevfileUtilsClient = NewDevfileUtilsClient()
```
maysunfaisal, line 168: so if we are setting it here does this mean, when HAS test calls `ParseDevfile()` it will not use the mock impl?

pkg/devfile/parser/parse.go
```diff
@@ -217,6 +165,11 @@ func ParseDevfile(args ParserArgs) (d DevfileObj, err error) {
...
+	if err != nil {
+		return d, fmt.Errorf("Error %v, URL cannot be parsed %s ", err, args.URL)
+	}
```
maysunfaisal, line 171: I think can be removed?

pkg/util/git.go
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package git
+package util
```
maysunfaisal, line 16: curious why we decided to move here in this pkg.. Seems like they could be in git pkg

pkg/util/util.go
```diff
@@ -1127,12 +1126,8 @@ func downloadInMemoryWithClient(params HTTPRequestParams, httpClient HTTPClient,
...
 			return nil, err
 		}
-		if !g.IsPublic(params.Timeout) {
-			// check that the token is valid before adding to the header
-			err = g.SetToken(params.Token, params.Timeout)
-			if err != nil {
-				return nil, err
-			}
+
+		if params.Token != "" {
```
maysunfaisal, line 1130: 👍

pkg/devfile/parser/parse.go
kim-tsao: Good catch, I will need to initialize it only when it's not initialized

pkg/util/git.go
kim-tsao: Originally, there were duplicates of the until functions in the git package because of cyclic dependencies.  It was confusing because some of the parser code was using the same named functions in the util package while others were using the ones from the git package.

pkg/devfile/parser/parse.go
kim-tsao: yes it looks like it was left there from another change that got removed

maysunfaisal (approved):

PR #186, opened 2023-10-24
Make the parser able to parse any file (including files with `.yml` extension)

rm3l:
### What does this PR do?:
This makes sure the Devfile parser does not make any assumptions on the path it is provided. It should now be able to parse any file path it is provided by clients, including files with a `.yml` extension.

After digging, I found no special place in the code where the library is discovering a Devfile.

Fixes https://github.com/devfile/api/issues/1265

### PR acceptance criteria:

- [x] Unit/Functional tests
- [x] Documentation (READMEs, Product Docs, Blogs, Education Modules, etc.)
- [x] Client Impact

I checked the following clients:
- `odo`: there is an ongoing PR (https://github.com/redhat-developer/odo/pull/7130). `odo` has a specified list of acceptable filenames for a Devfile, and this list has been extended to include `devfile.yml` and `.devfile.yml`
- `ODC`: they call the parser using the `Data` field, which should not be affected by the changes here (which handles parsing from a specified devfile path): https://github.com/openshift/console/blob/master/pkg/devfile/handler.go#L55. I tested based on the instructions [here](https://github.com/openshift/console?tab=readme-ov-file#openshift-console).
- Che: they have their own logic for parsing: [devfile/devworkspace-operator@main/pkg/library](https://github.com/devfile/devworkspace-operator/tree/main/pkg/library?rgh-link-date=2023-10-25T12%3A53%3A08Z)
So they should not be impacted by the changes here. Furthermore, they currently look by default for a devfile with the file name .devfile.yaml or devfile.yaml. For other cases, they allow users to specify a custom path, which also covers {,.}devfile.yml or any other files. I tested on Dev Sandbox using a devfile.yml file and it works without any problem.
See [eclipse.dev/che/docs/stable/end-user-guide/url-parameter-for-the-devfile-file-name](https://eclipse.dev/che/docs/stable/end-user-guide/url-parameter-for-the-devfile-file-name/) and [che-incubator/che-code@main/code/extensions/che-remote/src/extension.ts#L100-L112](https://github.com/che-incubator/che-code/blob/main/code/extensions/che-remote/src/extension.ts?rgh-link-date=2023-10-25T12%3A53%3A08Z#L100-L112)
- HAS: https://github.com/redhat-appstudio/application-service/pull/400 made it possible to allow `devfile.yml` and `.devfile.yml` as potential input devfiles. But parsing of those files currently fails. I checked with the changes here and parsing works as expected.
 
- [x] Gosec scans

- Parsing a context dir containing multiple devfiles. It should be able to detect and pick a devfile based on the following priority order: devfile.yaml > .devfile.yaml > devfile.yml > .devfile.yml
```shell
$ mkdir -p /tmp/test-parser
$ wget https://registry.devfile.io/devfiles/java-maven/1.3.0 -O /tmp/test-parser/devfile.yml

# Try to parse it. It should pass now.
# Previously, it would fail with the following message:
# failed to populateAndParseDevfile: 
#   failed to read devfile from path '/tmp/test-parser/devfile.yaml': 
#   open /tmp/test-parser/devfile.yaml: not a directory
$ go run . /tmp/test-parser

parsing devfile from /tmp/test-parser
schema version: 2.1.0
All component: 
tools
m2
All Exec commands: 
command mvn-package is with kind: build
workingDir is: ${PROJECT_SOURCE}
command run is with kind: run
workingDir is: ${PROJECT_SOURCE}
command debug is with kind: debug
workingDir is: ${PROJECT_SOURCE}
=========================================================
Container components applied filter: 
Exec commands applied filter: 
err: Attribute with key "alpha.build-dockerfile" does not exist
dockerfilePath: 
```

- Parsing any file, regardless of file name or extension

```shell
# Create devfile file on the local filesystem
# and name it however you want (e.g., devfile.yml)
$ wget https://registry.devfile.io/devfiles/go/1.0.2 -O /tmp/my-devfile.yml

# Try to parse it. It should pass now.
# Previously, it would fail with the following message:
# failed to populateAndParseDevfile: 
#   failed to read devfile from path '/tmp/devfile.yml/devfile.yaml': 
#   open /tmp/devfile.yml/devfile.yaml: not a directory
$ go run . /tmp/devfile.yml

parsing devfile from /tmp/devfile.yml
schema version: 2.1.0
All component: 
runtime
All Exec commands: 
command build is with kind: build
workingDir is: ${PROJECT_SOURCE}
command run is with kind: run
workingDir is: ${PROJECT_SOURCE}
=========================================================
Container components applied filter: 
Exec commands applied filter: 
err: Attribute with key "alpha.build-dockerfile" does not exist
dockerfilePath: 
```


README.md
```diff
@@ -8,11 +8,11 @@
 ## About
 
 The Devfile Parser library is a Golang module that:
-1. parses the devfile.yaml as specified by the [api](https://devfile.github.io/devfile/api-reference.html) & [schema](https://github.com/devfile/api/tree/main/schemas/latest).
-2. writes to the devfile.yaml with the updated data.
+1. parses a devfile as specified by the [api](https://devfile.github.io/devfile/api-reference.html) & [schema](https://github.com/devfile/api/tree/main/schemas/latest).
```
thepetk, line 11: > https://devfile.github.io/devfile/api-reference.html

The link returns 404. I think it should be updated

thepetk (commented):
I think the idea behind this solution makes sense.

One question I have is what about Che and HAS? Have we checked how they are first filtering the files passed to `ParseDevfileAndValidate`?

README.md
rm3l: Good catch. Where is the API reference located now? https://devfile.io/docs/2.2.1/devfile-schema?

rm3l: Fixed in [`44e7c7d` (#186)](https://github.com/devfile/library/pull/186/commits/44e7c7d4c142ba9886bee250397a82f17413fd35)

rm3l:
Whoops, I just noticed this comment..

> I think the idea behind this solution makes sense.
> 
> One question I have is what about Che and HAS? Have we checked how they are first filtering the files passed to `ParseDevfileAndValidate`?

I checked Che, and in my understanding, Che (the DevWorkspace Operator) has its own library for parsing Devfiles: https://github.com/devfile/devworkspace-operator/tree/main/pkg/library
So they should not be impacted by the changes here. Furthermore, they currently look by default for a devfile with the file name `.devfile.yaml` or `devfile.yaml`. For other cases, they allow users to specify a custom path, which also covers `{,.}devfile.yml` or any other files. I tested on Dev Sandbox using a `devfile.yml` file and it works without any problem.
See https://eclipse.dev/che/docs/stable/end-user-guide/url-parameter-for-the-devfile-file-name/ and https://github.com/che-incubator/che-code/blob/main/code/extensions/che-remote/src/extension.ts#L100-L112

This PR is more about making sure the parser is able to parse filesystem paths other than `devfile.yaml` and checking whether there are places where the library discovers devfiles, which does not seem to be the case.
Seeing that all tools seem to have their own logic for detecting default devfile file names (other than `devfile.yaml`) and that `.yml` is not supported by all tools, I was thinking of mentioning this point in the documentation (along with the recommended priority order if multiple variants of devfiles are present). I plan to do that in a subsequent PR.

As for HAS, I saw this has been covered too in DEVHAS-493 (https://github.com/redhat-appstudio/application-service/pull/400), where support for detecting `{,.}devfile.yml` has been added.
@yangcao77 Could you please confirm that on HAS?

rm3l:
I've gone ahead and tried to check the current changes on HAS.
I thought recognizing `[.]devfile.yml` was already implemented on HAS (via https://github.com/redhat-appstudio/application-service/pull/400), but when trying to run the unit tests under `cdq-analysis` ([`TestScanRepo`](https://github.com/redhat-appstudio/application-service/blob/main/cdq-analysis/pkg/devfile_test.go#L158)) against a multi-component repo with `devfile.yml` files (https://github.com/rm3l/rh-app-svc-devfile-test-multi-components-deep), I run into the same error as the one we are trying to address in this PR.

```
=== RUN   TestScanRepo
--- FAIL: TestScanRepo (1.21s)
=== RUN   TestScanRepo/Should_return_2_devfile_contexts,_and_2_devfileURLs_as_this_is_a_multi_comp_devfile
    devfile_test.go:297: 
got unexpected error invalid devfile due to err: 
failed to populateAndParseDevfile: 
failed to read devfile from path '/tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml/devfile.yaml': 
open /tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml/devfile.yaml: 
not a directory, 
failed to parse the devfile content from /tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml
    --- FAIL: TestScanRepo/Should_return_2_devfile_contexts,_and_2_devfileURLs_as_this_is_a_multi_comp_devfile (1.21s)
FAIL
```

(As I expected, the test passes with the changes in this PR)

@yangcao77 Am I missing anything about DEVHAS-493 and https://github.com/redhat-appstudio/application-service/pull/400? 

I guess HAS will need to update the Devfile library once the changes here are merged..

yangcao77:
the failure you saw is due to this:
https://github.com/devfile/library/blob/36607c5072830668282922b68178a30bf8fd8434/pkg/devfile/parser/context/context.go#L99-L101

in library, if the extension is not .yaml , it will append devfile.yaml at the end of the path. so HAS passed `/tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml`  to library and try to parse the content. then library thinks devfile.yml  is not supported devfile extention, and append devfile.yaml .
so you see the error message:
```
failed to read devfile from path '/tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml/devfile.yaml': open /tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml/devfile.yaml: not a directory, failed to parse the devfile content from /tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml
```

I saw in your PR, you removed the block which appends devfile.yaml , it will fix the current issue, but I’m not suggesting to do it. it will cause regression to existing users who only pass the context  path to library

and also to test with your changes in library with HAS and ODC(please do), need to replace the library version in go.mod
```
replace github.com/devfile/library/v2 => https://github.com/rm3l/devfile-library <the commit go version>
```

yangcao77:
> I guess HAS will need to update its version of the Devfile library once the changes here are merged..

Yes, HAS need the change in library to be able to parse from the content.

pkg/devfile/parser/context/context.go
```diff
@@ -96,17 +92,6 @@ func (d *DevfileCtx) populateDevfile() (err error) {
 
 // Populate fills the DevfileCtx struct with relevant context info
 func (d *DevfileCtx) Populate() (err error) {
```
yangcao77, line 94: should not remote this, this will cause regression to users who only pass in the context path to library.

pkg/devfile/parse_test.go
```diff
@@ -527,6 +527,90 @@ spec:
 				StarterProjects: map[string][]string{},
 			},
 		},
+		{
```
yangcao77, line 530: perhaps add another test to test with context `./testdata/`, this was be supported by previous versions of library, need to ensure it's backward-compatible. 

the priority should be `devfile.yaml` > `.devfile.yaml` > `devfile.yml` > `.devfile.yml`

pkg/devfile/parser/writer_test.go
```diff
@@ -17,67 +17,96 @@ package parser
...
+		{
+			name:     "write devfile with .yml extension",
+			fileName: "devfile.yml",
+		},
+		{
+			name:     "write .devfile with .yml extension",
+			fileName: ".devfile.yml",
+		},
+		{
+			name:     "write any file, regardless of name and extension",
```
yangcao77, line 61: this is out-of-scope for the current issue

yangcao77 (changes requested):

pkg/devfile/parser/context/context.go
rm3l: You mean users passing the parent directory containing the devfile(s)? Do you know which clients are passing a directory instead of the actual devfile path? The ones that I tested (`odo`, ODC) are parsing either from a file path (`odo`) or from in-memory bytes ([ODC](https://github.com/openshift/console/blob/master/pkg/devfile/handler.go#L55)).
And from the [`ParserArgs#Path`](https://github.com/devfile/library/blob/main/pkg/devfile/parser/parse.go#L83-L84) field description, it seems that we are expecting a relative or absolute path to the devfile path, right?
Anyway, just to be safe, I'll change that to make sure both cases are handled. Thanks!

pkg/devfile/parser/writer_test.go
rm3l: Actually, part of the acceptance criteria is for the library to work generically and support any file name (not just `.yml`)..

rm3l:
> the failure you saw is due to this:
> 
> https://github.com/devfile/library/blob/36607c5072830668282922b68178a30bf8fd8434/pkg/devfile/parser/context/context.go#L99-L101
> 
> in library, if the extension is not .yaml , it will append devfile.yaml at the end of the path. so HAS passed `/tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml` to library and try to parse the content. then library thinks devfile.yml is not supported devfile extention, and append devfile.yaml . so you see the error message:
> 
> ```
> failed to read devfile from path '/tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml/devfile.yaml': open /tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml/devfile.yaml: not a directory, failed to parse the devfile content from /tmp/testclone/devfile-sample-java-springboot-basic/.devfile/devfile.yml
> ```
> 
> I saw in your PR, you removed the block which appends devfile.yaml , it will fix the current issue, but I’m not suggesting to do it. it will cause regression to existing users who only pass the context path to library
> 
https://github.com/devfile/library/pull/186#discussion_r1377756837

> and also to test with your changes in library with HAS and ODC(please do), need to replace the library version in go.mod
> 
> ```
> replace github.com/devfile/library/v2 => https://github.com/rm3l/devfile-library <the commit go version>
> ```

Yes, that's how I am testing this with `odo` for example (with `go mod edit -replace ...`).

yangcao77 (approved):
change looks good to me.  assumes verified it works with consumers

rm3l:
> change looks good to me. assumes verified it works with consumers

Thanks for the review!
Yes, I manually checked that this works with consumers and updated the [issue description](https://github.com/devfile/library/pull/186#issue-1958948793) accordingly for reference. For ODC specifically, I manually tested the changes here based on the instructions [here](https://github.com/openshift/console?tab=readme-ov-file#openshift-console).

PR #208, opened 2024-03-14
add support for 2.2.1 and 2.2.2 schema versions

Jdubrick:
### What does this PR do?:

A summary of changes made in this PR can be found below:
1. Added schema version 2.2.1 and 2.2.2 to the library, they are now recognized and parsable.
2. Updated 2.2.0 version to reference 2.2.0 release instead of 2.1.1-alpha as 2.1.1 is now added itself
3. Added new case to existing unit test to verify unsupported schemas are being picked up and tested

fixes https://github.com/devfile/api/issues/1454

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Client Impact

  

- [ ] Gosec scans
  

Unit tests ran with `make test` should pass. Additionally when you try and parse a devfile on schema 2.2.1 or 2.2.2 now it should function and not spit an error message.

### DIFF
--- a/pkg/devfile/parse_test.go
+++ b/pkg/devfile/parse_test.go
@@ -152,7 +152,7 @@ metadata:
   tags:
   - Go
   version: 1.0.0
-schemaVersion: 2.2.0
+schemaVersion: 2.2.2
 `
 
 	devfileContentWithVariable := devfileContent + `variables:
@@ -267,6 +267,8 @@ spec:
 	testServer.Start()
 	defer testServer.Close()
 
+	unsupportedSchemaError := `error parsing devfile because of non-compliant data due to unable to find schema for version "2.2.5". The parser supports devfile schema for version 2.0.0, 2.1.0, 2.2.0, 2.2.1, 2.2.2, v1alpha2`
+
 	type args struct {
 		args parser.ParserArgs
 	}
@@ -279,7 +281,7 @@ spec:
 		wantOpenshiftInline  string
 		wantVariables        map[string]string
 		additionalChecks     func(parser.DevfileObj) error
-		wantError            bool
+		wantErrorStr         *string
 	}{
 		{
 			name: "with external overriding variables",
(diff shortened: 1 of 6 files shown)


pkg/devfile/parse_test.go
```diff
@@ -751,12 +775,38 @@ spec:
...
+			},
+			wantError: true,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			gotD, gotVarWarning, err := ParseDevfileAndValidate(tt.args.args)
-			if err != nil {
+			if err != nil && tt.wantError {
+				t.Log("Correctly identified unsupported schema version")
```
michael-valdron, line 807: In the event we want to support other expected errors, could we add something like `wantErrStr` to see which error should be expected?

pkg/devfile/parse_test.go
```diff
@@ -751,12 +775,38 @@ spec:
...
+				return
+			} else if err != nil {
 				t.Errorf("ParseDevfileAndValidate() error = %v, wantErr nil", err)
```
michael-valdron, line 810: Not part of your changes but just a nit pick correction to the current source code:
```suggestion
				t.Errorf("ParseDevfileAndValidate() error = %v, wantErr false", err)
```

Or maybe with my other suggestion:
```suggestion
				t.Errorf("ParseDevfileAndValidate() error = %v, wantErrStr = %v", err, tt.wantErrStr)
```

michael-valdron: If using `wantErrStr` maybe typed as `*string` would show `nil` in cases where it is unset.

pkg/devfile/parse_test.go
Jdubrick: Yeah I agree `wantErrStr` would be more appropriate here so we can verify the correct errors are occurring as well. Let me make some changes to it

Jdubrick:
@michael-valdron i pushed a new commit for the changes. I took your advice about having the string pointer so if it needs an error it sets the string and if not it will be nil. I also had to make sure the schema versions were sorted so the output was in the correct order for error messages, im pretty sure this is due to the schemas being stored in a map so they're all out of order.

michael-valdron (approved):
/lgtm

PR #213, opened 2024-06-05
Update schema version to 2.2.2 for nodejs test child devfiles

michael-valdron:
### What does this PR do?:

Fixes mock child devfiles of the `nodejs` stack by bumping the schema version to `2.2.2`.

### Which issue(s) this PR fixes:

fixes devfile/api#1589

### PR acceptance criteria:
Testing and documentation do not need to be complete in order for this PR to be approved. We just need to ensure tracking issues are opened and linked to this PR, if they are not in the PR scope due to various constraints.

- [X] Unit/Functional tests

  

- [X] [QE Integration test](https://github.com/devfile/integration-tests) 

  

- [ ] Documentation (READMEs, Product Docs, Blogs, Education Modules, etc.)

   

- [ ] Client Impact

  

- [ ] Gosec scans
  

### How to test changes / Special notes to the reviewer:

```
make test
```

PR #216, opened 2024-06-11
Fix parent version on test devfile

thepetk:
### What does this PR do?:
The `Test_Parent_RegistryURL.yaml` consumes the latest parent version of `nodejs`. As a result any breaking changes will impact the library's tests. This PR simply fixes the version to `2.2.1` (current latest) so the test related to this file remain unimpacted by any changes of the `devfile/registry)

N/A

### PR acceptance criteria:

- [ ] Unit/Functional tests

  

- [ ] Client Impact

  

- [ ] Gosec scans

### DIFF
--- a/tests/v2/devfiles/samples/Test_Parent_RegistryURL.yaml
+++ b/tests/v2/devfiles/samples/Test_Parent_RegistryURL.yaml
@@ -2,6 +2,7 @@ schemaVersion: 2.2.2
 parent:
   id: nodejs
   registryUrl: "https://registry.stage.devfile.io"
+  version: 2.2.0
   commands:
   - id: run
     exec:
--- a/tests/v2/devfiles/samples/Test_Parent_RegistryURL.yaml
+++ b/tests/v2/devfiles/samples/Test_Parent_RegistryURL.yaml
@@ -2,7 +2,7 @@ schemaVersion: 2.2.2
 parent:
   id: nodejs
   registryUrl: "https://registry.stage.devfile.io"
-  version: 2.2.0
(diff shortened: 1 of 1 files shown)


Jdubrick (commented):
This looks like it is reverting changes made [here](https://github.com/devfile/library/pull/213). Do we need to change the other items in that PR as well?

thepetk:
> This looks like it is reverting changes made [here](https://github.com/devfile/library/pull/213). Do we need to change the other items in that PR as well?

No I don't think so, I didn't notice this change. I think the PR I'm mentioning needs to be rebased instead.

thepetk:
> > This looks like it is reverting changes made [here](https://github.com/devfile/library/pull/213). Do we need to change the other items in that PR as well?
> 
> No I don't think so, I didn't notice this change. I think the PR I'm mentioning needs to be rebased instead.

@Jdubrick although I think might worth pin the `parent` version to `2.2.1` so we don't have the same issue again when someone updates the nodejs schemaversion.

cc: @michael-valdron

Jdubrick:
2.2.1 sounds good to me. I think we should open an issue for overhauling the tests so we can properly test 2.2.2 schemas, wdyt? @thepetk

thepetk:
> 2.2.1 sounds good to me. I think we should open an issue for overhauling the tests so we can properly test 2.2.2 schemas, wdyt? @thepetk

Yeah I didn't open something cause @michael-valdron has the intention to bring it up as a discussion topic. If needed I'll create one afterwards :)

Jdubrick (approved):
/lgtm

PR #220, opened 2024-08-03
add projectsrc to env

Horiodino:
### What does this PR do?:
Updating  ENV list for all special/reserved ENVs

fixes: https://github.com/devfile/api/issues/1601

### PR acceptance criteria:

- [x] Unit/Functional tests

  

- [ ] Client Impact

  

- [ ] Gosec scans

### DIFF
--- a/pkg/devfile/generator/generators_test.go
+++ b/pkg/devfile/generator/generators_test.go
@@ -123,6 +123,16 @@ func TestGetContainers(t *testing.T) {
 							Container: v1.Container{
 								Image:        containerImages[0],
 								MountSources: &trueMountSources,
+								Env: []v1.EnvVar{
+									{
+										Name:  "testVar1",
+										Value: "testVal1",
+									},
+									{
+										Name:  "testVar2",
+										Value: "testVal2",
+									},
+								},
 							},
 						},
 					},
@@ -132,12 +142,20 @@ func TestGetContainers(t *testing.T) {
 			wantContainerImage: containerImages[0],
 			wantContainerEnv: []corev1.EnvVar{
 				{
+					Name:  "PROJECT_SOURCE",
+					Value: "/projects/test-project",
+				},
+				{
 					Name:  "PROJECTS_ROOT",
 					Value: "/projects",
 				},
 				{
-					Name:  "PROJECT_SOURCE",
-					Value: "/projects/test-project",
+					Name:  "testVar1",
+					Value: "testVal1",
+				},
+				{
+					Name:  "testVar2",
+					Value: "testVal2",
 				},
 			},
 			wantContainerVolMount: []corev1.VolumeMount{
@@ -158,6 +176,16 @@ func TestGetContainers(t *testing.T) {
 								Image:         containerImages[0],
 								MountSources:  &trueMountSources,
 								SourceMapping: "/myroot",
+								Env: []v1.EnvVar{
+									{
+										Name:  "testVar1",
+										Value: "testVal1",
+									},
+									{
+										Name:  "testVar2",
+										Value: "testVal2",
+									},
+								},
 							},
 						},
 					},
@@ -167,16 +195,20 @@ func TestGetContainers(t *testing.T) {
 			wantContainerImage: containerImages[0],
 			wantContainerEnv: []corev1.EnvVar{
 				{
+					Name:  "PROJECT_SOURCE",
+					Value: "/myroot/test-project",
+				},
+				{
 					Name:  "PROJECTS_ROOT",
 					Value: "/myroot",
 				},
 				{
-					Name:  "PROJECT_SOURCE",
-					Value: "/myroot",
+					Name:  "testVar1",
+					Value: "testVal1",
 				},
 				{
-					Name:  "PROJECT_SOURCE",
-					Value: "/myroot/test-project",
+					Name:  "testVar2",
+					Value: "testVal2",
 				},
 			},
 			wantContainerVolMount: []corev1.VolumeMount{
--- a/pkg/devfile/generator/generators_test.go
+++ b/pkg/devfile/generator/generators_test.go
@@ -52,7 +52,6 @@ func init() {
 }
 
 func TestGetContainers(t *testing.T) {
-
 	containerNames := []string{"testcontainer1", "testcontainer2", "testcontainer3"}
 	containerImages := []string{"image1", "image2", "image3"}
 
@@ -132,9 +131,12 @@ func TestGetContainers(t *testing.T) {
 			wantContainerName:  containerNames[0],
 			wantContainerImage: containerImages[0],
 			wantContainerEnv: []corev1.EnvVar{
-
 				{
 					Name:  "PROJECTS_ROOT",
+					Value: "/projects",
+				},
+				{
+					Name:  "PROJECT_SOURCE",
 					Value: "/projects",
 				},
 				{
@@ -168,9 +170,12 @@ func TestGetContainers(t *testing.T) {
 			wantContainerName:  containerNames[0],
 			wantContainerImage: containerImages[0],
 			wantContainerEnv: []corev1.EnvVar{
-
 				{
 					Name:  "PROJECTS_ROOT",
+					Value: "/myroot",
+				},
+				{
+					Name:  "PROJECT_SOURCE",
 					Value: "/myroot",
 				},
 				{
@@ -309,7 +314,6 @@ func TestGetContainers(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-
 			ctrl := gomock.NewController(t)
 			defer ctrl.Finish()
 			mockDevfileData := data.NewMockDevfileData(ctrl)
@@ -372,7 +376,6 @@ func TestGetContainers(t *testing.T) {
 			}
 		})
 	}
-
 }
 
 func TestGetVolumesAndVolumeMounts(t *testing.T) {
@@ -1449,6 +1452,7 @@ func TestGetPodTemplateSpec(t *testing.T) {
 							Env: []corev1.EnvVar{
 								{Name: "PROJECTS_ROOT", Value: "/projects"},
 								{Name: "PROJECT_SOURCE", Value: "/projects"},
+								{Name: "PROJECT_SOURCE", Value: "/projects"},
 							},
 							ImagePullPolicy: corev1.PullAlways,
 							SecurityContext: &corev1.SecurityContext{
@@ -1460,6 +1464,7 @@ func TestGetPodTemplateSpec(t *testing.T) {
 							Image: "a-tool-image",
 							Env: []corev1.EnvVar{
 								{Name: "PROJECTS_ROOT", Value: "/projects"},
+								{Name: "PROJECT_SOURCE", Value: "/projects"},
 								{Name: "PROJECT_SOURCE", Value: "/projects"},
 							},
 							ImagePullPolicy: corev1.PullAlways,
@@ -1512,6 +1517,7 @@ func TestGetPodTemplateSpec(t *testing.T) {
 							Image: "an-image",
 							Env: []corev1.EnvVar{
 								{Name: "PROJECTS_ROOT", Value: "/projects"},
+								{Name: "PROJECT_SOURCE", Value: "/projects"},
 								{Name: "PROJECT_SOURCE", Value: "/projects"},
 							},
 							ImagePullPolicy: corev1.PullAlways,
@@ -1566,6 +1572,7 @@ func TestGetPodTemplateSpec(t *testing.T) {
 							Image: "an-image",
 							Env: []corev1.EnvVar{
 								{Name: "PROJECTS_ROOT", Value: "/projects"},
+								{Name: "PROJECT_SOURCE", Value: "/projects"},
 								{Name: "PROJECT_SOURCE", Value: "/projects"},
 							},
 							ImagePullPolicy: corev1.PullAlways,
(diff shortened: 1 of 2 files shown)


Horiodino:
shoudld i remove `addSyncFolder` module ? as the `EnvProjectsSrc` repeated two times

pkg/devfile/generator/generators_test.go
```diff
@@ -168,11 +170,14 @@ func TestGetContainers(t *testing.T) {
 			wantContainerName:  containerNames[0],
 			wantContainerImage: containerImages[0],
 			wantContainerEnv: []corev1.EnvVar{
-
 				{
 					Name:  "PROJECTS_ROOT",
 					Value: "/myroot",
 				},
+				{
+					Name:  "PROJECT_SOURCE",
```
Jdubrick, line 178: Why are we adding `PROJECT_SOURCE` here twice with different values?

Jdubrick: I believe the issue is that the value is not being read properly if it is the final item but if that's the case I think we could probably find a better solution than just duplicating the value. wdyt @michael-valdron @thepetk am I understanding this correctly?

Horiodino:
i was thinking maybe rather than using `addSyncFolder` why wont just merge that module with `addSyncRootFolder`  ? 
does the same work ?

thepetk (commented):
Apologies, I just realized the review I left on Wednesday left on "pending" state.

pkg/devfile/generator/utils.go
```diff
@@ -157,12 +157,17 @@ func addSyncRootFolder(container *corev1.Container, sourceMapping string) string
...
-			Name:  EnvProjectsSrc,
-			Value: syncRootFolder,
 		},
 	}, container.Env...)
 
+	if sourceMapping != "" {
+		container.Env = append(container.Env,
+			corev1.EnvVar{
+				Name:  EnvProjectsSrc,
+				Value: sourceMapping,
```
Jdubrick, line 167: Out of curiosity, why do we only want to set `EnvProjectsSrc` if `sourceMapping` is set? Since if `sourceMapping` is not set then we assign `syncRootFolder = DevfileSourceVolumeMount` is that not the source we would need? Not saying it is wrong I just want to understand how it is working because if I am not mistaken we aren't setting `EnvProjectsSrc` if `sourceMapping` is unset

Horiodino: `I think here it should take the source value not the root.`
Ah, I implemented it as described, but I think I misunderstood something.

Jdubrick: @thepetk do you have context related to your comment here? https://github.com/devfile/library/pull/220#discussion_r1706961576

Horiodino: the main problem is the we get the same identical values ? which should not be there right ? we doesn't append if the values are same am i correct ?

thepetk: > @thepetk do you have context related to your comment here? https://github.com/devfile/library/pull/220#discussion_r1706961576

From what I see, we're assigning everytime the root dir as project source. However I'd expect source to be `PROJECT_ROOT` + `RELATIVE_SOURCE_PATH`.

Now, following the flow of the library both source & root are set [here](https://github.com/devfile/library/blob/main/pkg/devfile/generator/utils.go#L719). So, in case no project has been found there's no need for `PROJECT_SOURCE` path. In case we'd like to set `PROJECT_SOURCE` no matter what I'd expect updating (suggestion):

1. The [addSyncFolder](https://github.com/devfile/library/blob/main/pkg/devfile/generator/utils.go#L168-L199) function to include this case (inc test cases updates).
2. The [getAllContainers](https://github.com/devfile/library/blob/main/pkg/devfile/generator/utils.go#L689-L734) function by removing the early return for no-projects-found cases and this way we set `PROJECT_SOURCE` everytime.

My main question here is if it's ok to add a value for `PROJECT_SOURCE` even if we haven't found anything. That said I'd expect a null value for this env var for such cases.

pkg/devfile/generator/utils.go
```diff
@@ -153,11 +153,20 @@ func addSyncRootFolder(container *corev1.Container, sourceMapping string) string
...
 	// Add PROJECTS_ROOT to the container
-	container.Env = append(container.Env,
-		corev1.EnvVar{
+	container.Env = append([]corev1.EnvVar{
+		{
 			Name:  EnvProjectsRoot,
 			Value: syncRootFolder,
-		})
+		},
+	}, container.Env...)
```
Jdubrick, line 161: I was looking into the issue. The goal of the issue associated with this PR is to ensure that `PROJECT_ROOT` and `PROJECT_SOURCE` are the first 2 environment variables  in the env var list. The way this append is structured with the spread of `container.Env...` after should essentially be prepending the value to the front of the slice. I believe if we do that for both `PROJECT_ROOT` and `PROJECT_SOURCE` then we are good @thepetk

thepetk: Oh ok that makes sense. So my comment based on this, it might be better to check that the `PROJECT_ROOT` is on the top of the env vars list in [addSyncRootFolder](https://github.com/devfile/library/blob/64f49c75c68f6daf5e1b1cb26e2acd40eb57f8db/pkg/devfile/generator/utils.go#L146) (we're already doing that) and the `PROJECT_SOURCE` in the [addSyncFolder](https://github.com/devfile/library/blob/main/pkg/devfile/generator/utils.go#L168-L199).

That said I think this [block](https://github.com/devfile/library/blob/64f49c75c68f6daf5e1b1cb26e2acd40eb57f8db/pkg/devfile/generator/utils.go#L163-L169) should be removed, while this [block](https://github.com/devfile/library/blob/64f49c75c68f6daf5e1b1cb26e2acd40eb57f8db/pkg/devfile/generator/utils.go#L201-L205) should make sure that the env var is on the top of the env vars list, just like we're doing here.

pkg/devfile/generator/generators_test.go
thepetk: Better to resolve in favor of https://github.com/devfile/library/pull/220/files#r1717510608

pkg/devfile/generator/utils.go
thepetk: Better to resolve in favor of https://github.com/devfile/library/pull/220/files#r1717510608 and we can continue the discussion there :)

pkg/devfile/generator/generators_test.go
```diff
@@ -168,11 +166,14 @@ func TestGetContainers(t *testing.T) {
...
```
Jdubrick, line 174:
```suggestion
					Name:  "PROJECT_SOURCE",
```
We don't need this `PROJECT_SOURCE`

pkg/devfile/generator/utils.go
```diff
@@ -153,11 +153,20 @@ func addSyncRootFolder(container *corev1.Container, sourceMapping string) string
...
+
+	if sourceMapping != "" {
```
Jdubrick, line 163: We don't need this check in `addSyncRootFolder` we only need to add `EnvProjectsRoot` which you are doing above.

Jdubrick (changes requested):
After taking a look into this and testing it myself the following changes should be made in addition to my review comments:

In [`addSyncFolder`](https://github.com/Horiodino/library/blob/main/pkg/devfile/generator/utils.go#L168) we should turn this:
```
container.Env = append(container.Env,
		corev1.EnvVar{
			Name:  EnvProjectsSrc,
			Value: syncFolder,
		})
```
Into this:
```
container.Env = append([]corev1.EnvVar{
		{
			Name:  EnvProjectsSrc,
			Value: syncFolder,
		},
	}, container.Env...)
```

We can also update [`TestGetContainers`](https://github.com/Horiodino/library/blob/main/pkg/devfile/generator/generators_test.go#L118-#L188) test cases to the following to observe the ordering:
```
{
			name: "Container with default project root",
			containerComponents: []v1.Component{
				{
					Name: containerNames[0],
					ComponentUnion: v1.ComponentUnion{
						Container: &v1.ContainerComponent{
							Container: v1.Container{
								Image:        containerImages[0],
								MountSources: &trueMountSources,
								Env: []v1.EnvVar{
									{
										Name:  "testVar1",
										Value: "testVal1",
									},
									{
										Name:  "testVar2",
										Value: "testVal2",
									},
								},
							},
						},
					},
				},
			},
			wantContainerName:  containerNames[0],
			wantContainerImage: containerImages[0],
			wantContainerEnv: []corev1.EnvVar{
				{
					Name:  "PROJECT_SOURCE",
					Value: "/projects/test-project",
				},
				{
					Name:  "PROJECTS_ROOT",
					Value: "/projects",
				},
				{
					Name:  "testVar1",
					Value: "testVal1",
				},
				{
					Name:  "testVar2",
					Value: "testVal2",
				},
			},
			wantContainerVolMount: []corev1.VolumeMount{
				{
					Name:      "devfile-projects",
					MountPath: "/projects",
				},
			},
		},
		{
			name: "Container with source mapping",
			containerComponents: []v1.Component{
				{
					Name: containerNames[0],
					ComponentUnion: v1.ComponentUnion{
						Container: &v1.ContainerComponent{
							Container: v1.Container{
								Image:         containerImages[0],
								MountSources:  &trueMountSources,
								SourceMapping: "/myroot",
								Env: []v1.EnvVar{
									{
										Name:  "testVar1",
										Value: "testVal1",
									},
									{
										Name:  "testVar2",
										Value: "testVal2",
									},
								},
							},
						},
					},
				},
			},
			wantContainerName:  containerNames[0],
			wantContainerImage: containerImages[0],
			wantContainerEnv: []corev1.EnvVar{
				{
					Name:  "PROJECT_SOURCE",
					Value: "/myroot/test-project",
				},
				{
					Name:  "PROJECTS_ROOT",
					Value: "/myroot",
				},
				{
					Name:  "testVar1",
					Value: "testVal1",
				},
				{
					Name:  "testVar2",
					Value: "testVal2",
				},
			},
			wantContainerVolMount: []corev1.VolumeMount{
				{
					Name:      "devfile-projects",
					MountPath: "/myroot",
				},
			},
		},
```

Jdubrick:
fyi @thepetk if you see my review and want to try the changes yourself

pkg/devfile/generator/utils.go
Horiodino: > I was looking into the issue. The goal of the issue associated with this PR is to ensure that `PROJECT_ROOT` and `PROJECT_SOURCE` are the first 2 environment variables in the env var list. The way this append is structured with the spread of `container.Env...` after should essentially be prepending the value to the front of the slice. I believe if we do that for both `PROJECT_ROOT` and `PROJECT_SOURCE` then we are good @thepetk

but currently `PROJECT_SOURCE` taking same value as `PROJECT_ROOT` . i was thinking as you said append the two first but we shouldnt appedn blindly , we can just check if the both have same values then it should be null .

thepetk:
> fyi @thepetk if you see my review and want to try the changes yourself

@Jdubrick yes +1 is aligned with my comment [here](https://github.com/devfile/library/pull/220#discussion_r1718547995)

pkg/devfile/generator/utils.go
thepetk: @Horiodino based on the comment I made above and @Jdubrick requested changes, I think our suggested workaround is not changing the way we append the env vars, it only makes sure that each time we append `PROJECT_SOURCE` and/or `PROJECT_ROOT` we make sure they are on the top.

pkg/devfile/generator/utils.go
```diff
@@ -168,13 +169,11 @@ func addSyncRootFolder(container *corev1.Container, sourceMapping string) string
 func addSyncFolder(container *corev1.Container, sourceVolumePath string, projects []v1.Project) error {
 	var syncFolder string
 
-	// if there are no projects in the devfile, source would be synced to $PROJECTS_ROOT
 	if len(projects) == 0 {
-		syncFolder = sourceVolumePath
+		// No projects found, set PROJECT_SOURCE to empty
+		syncFolder = ""
```
Jdubrick, line 174: I don't believe we need any changes here, the only change that is required is the `container.Env` change you made below to insert the special env at the start of the slice. 

In addition I think it may be best to revert and include all the deleted comments, reviewing this issue was slightly confusing so it may be good to over document this portion.

Jdubrick (changes requested):

Jdubrick:
@Horiodino I believe the tests Go tests are failing because the order of the env vars in the tests is 1: PROJECT_ROOT and 2: PROJECT_SOURCE but since we are prepending and `PROJECT_SOURCE` is added last it is now coming first. We will need to adjust these tests. You should be able to reference the failing tests [here](https://github.com/devfile/library/actions/runs/10438447749/job/28948116388?pr=220)

Horiodino:
```--- PASS: Test_VolumeComponent (3.04s)
=== RUN   Test_VolumeComponentEdit
--- PASS: Test_VolumeComponentEdit (3.03s)
=== RUN   Test_MultiComponent
--- PASS: Test_MultiComponent (3.09s)
=== RUN   Test_Projects
--- PASS: Test_Projects (3.06s)
=== RUN   Test_StarterProjects
--- PASS: Test_StarterProjects (3.10s)
=== RUN   Test_Events
--- PASS: Test_Events (3.05s)
=== RUN   Test_EventsEdit
--- PASS: Test_EventsEdit (3.22s)
=== RUN   Test_Metadata
--- PASS: Test_Metadata (3.02s)
=== RUN   Test_MetadataEdit
--- PASS: Test_MetadataEdit (3.07s)
=== RUN   Test_Parent_Local_URI
--- PASS: Test_Parent_Local_URI (3.02s)
=== RUN   Test_v200_Devfile
--- PASS: Test_v200_Devfile (3.05s)
=== RUN   Test_v210_Devfile
--- PASS: Test_v210_Devfile (3.01s)
=== RUN   Test_v220_Devfile
--- PASS: Test_v220_Devfile (0.01s)
=== RUN   Test_Parent_KubeCRD
--- PASS: Test_Parent_KubeCRD (3.01s)
=== RUN   Test_Parent_RegistryURL
--- PASS: Test_Parent_RegistryURL (6.85s)
=== RUN   Test_Everything
--- PASS: Test_Everything (3.06s)
=== RUN   Test_EverythingEdit
--- PASS: Test_EverythingEdit (3.11s)
PASS
coverage: [no statements]
ok  	github.com/devfile/library/v2/tests/v2/libraryTest	92.911s	coverage: [no statements]
```

Jdubrick (approved):
/lgtm

@thepetk I know you reviewed alongside me if you want to also give a final review :)

